PR #8, opened 2019-06-17
lmp/jobserv: add builds for raspberrypi-cm3

mike-scott:
Due to recent subscriber interested, let's enable RPi3 computer module
builds (32-bit for now).

### DIFF
--- a/zephyr/jobserv.yml
+++ b/zephyr/jobserv.yml
@@ -84,10 +84,16 @@ triggers:
 params:
   archive: /archive
   FLAKY_TESTS: |
+    samples/drivers/led_lp5562
+    samples/subsys/usb/cdc_acm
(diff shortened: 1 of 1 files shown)


lmp/jobserv.yml
```diff
@@ -52,6 +52,7 @@ triggers:
               - intel-corei7-64
               - hikey
               - raspberrypi0-wifi
+              - raspberrypi-cm3
```
doanac: I don't think we can add this here, because we don't have a CI worker for this type.

mike-scott: You're right.  I misunderstood the comment for that section.  Fixed.

ricardosalveti:
Please add for both ci and post merge.

PR #9, opened 2019-06-18
lmp: replace 2 occurrences of zmp-sdk with lmp-sdk

mike-scott:


mike-scott:
@doanac noticed this while looking at zmp breakage.  Doesn't seem like this should be working as it is currently.

doanac:
We should probably just eliminate the lava testing here or move them to jobserv. However, they should use the zmp-sdk. Its triggering a lava job and needs tools from that container that aren't in lmp-sdk if i remember correctly.

mike-scott:
@doanac ok thanks for the clarification.  Closing.

PR #19, opened 2019-07-17
Add scripts for "factory containers"

doanac:
These are the scripts I've been using for building multi-arch containers in CI and then having their corresponding docker-app(s) published into an OTA targets.json.

### DIFF
--- a/factory-containers/build.sh
+++ b/factory-containers/build.sh
@@ -1,4 +1,5 @@
 #!/bin/sh -e
+# Copyright (c) 2019 Foundries.io, SPDX-License-Identifier: Apache-2.0
 set -o pipefail
 
 HERE=$(dirname $(readlink -f $0))
--- a/factory-containers/build.sh
+++ b/factory-containers/build.sh
@@ -1,0 +1,108 @@
+#!/bin/sh -e
+set -o pipefail
+
+HERE=$(dirname $(readlink -f $0))
+source $HERE/../helpers.sh
+
+require_params FACTORY
+
+run apk --no-cache add file git
+
+ARCH=amd64
+file /bin/busybox | grep -q aarch64 && ARCH=arm64 || true
+file /bin/busybox | grep -q armhf && ARCH=arm || true
+
+manifest_arch=$ARCH
+[ $manifest_arch = "arm" ] && manifest_arch=armv6
+status Grabbing manifest-tool for $manifest_arch
+wget -O /bin/manifest-tool https://github.com/estesp/manifest-tool/releases/download/v0.8.0/manifest-tool-linux-$manifest_arch
+chmod +x /bin/manifest-tool
+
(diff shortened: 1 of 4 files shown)


doanac:
Here's an example of it in action: https://ci.foundries.io/projects/andy-corp/lmp/builds/86/

factory-containers/build.sh
```diff
@@ -0,0 +1,108 @@
+#!/bin/sh -e
```
ricardosalveti: $ checkbashisms factory-containers/build.sh
possible bashism in factory-containers/build.sh line 5 (should be '.', not 'source'):
source $HERE/../helpers.sh
possible bashism in factory-containers/build.sh line 40 (${foo:3[:1]}):
TAG=${GIT_SHA:0:6}

If not changed to be generic, maybe force bash instead?

factory-containers/docker-app-publish.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/sh -e
```
ricardosalveti: $ checkbashisms factory-containers/docker-app-publish.sh
possible bashism in factory-containers/docker-app-publish.sh line 5 (should be '.', not 'source'):
source $HERE/../helpers.sh
possible bashism in factory-containers/docker-app-publish.sh line 10 (${foo:3[:1]}):
TAG=${GIT_SHA:0:6}
possible bashism in factory-containers/docker-app-publish.sh line 25 (alternative test command ([[ foo ]] should be [ foo ])):
if [[ "$sha" = "$newsha" ]] && [[ -n "$apps" ]] ; then

Same here.

factory-containers/test_build.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash -eu
```
ricardosalveti: This one doesn't need to use bash directly :-)

ricardosalveti:
Mind also adding: Copyright (c) 2019 Foundries.io, SPDX-License-Identifier: Apache-2.0 to the .sh files?

factory-containers/build.sh
doanac: its got to bin /bin/sh - docker:dind doesn't have bash.

doanac:
I think the comments are resolved. I'm testing the actual code here: 
  https://ci.foundries.io/projects/andy-corp/lmp/builds/88/

doanac:
i decided to keep with manifest-tool for now. Its a bit more invasive to get into our code base at the moment. We should look at just moving the whole repo over to it in the future.

ricardosalveti:
> i decided to keep with manifest-tool for now. Its a bit more invasive to get into our code base at the moment. We should look at just moving the whole repo over to it in the future.

Sounds good.

LGTM, just not so familiar with the codebase for more in depth review.

PR #22, opened 2019-08-13
lmp: Add build for aktualizr-lite target

doanac:


doanac:
FYI - this is what I'm thinking for the new "lite" builds. I'm going to generate a new stream on our ota-lite deployment called "public" and use the credentials from it for publishing. I think this would then give us an anonymous stream people could play with.

PR #23, opened 2019-08-29
lmp: Move to using aktualizr-lite

doanac:
This changes our CI process to build all targets for aktualizr-lite and then test them with it as well.


doanac:
There's some release promotion logic left, but that's in our private repo.

PR #24, opened 2019-09-06
Publish license manifest and required source tarball

ricardosalveti:


ricardosalveti:
> Looks fine. That's enough shell trickery that I'll trust you. I don't suppose there's a point to making this something we can disable via an environment variable? ie - who wouldn't want/need this?

Right, it should be enabled by default for all public builds and at the moment we don't have multiple streams for subscribers (premerge/postmerge), so I would say better to have it always on by default.

Once we get multiple streams we can probably just enabled for the postmerge/release one.

PR #28, opened 2019-09-30
lmp/build.sh: support EXTRA_ARTIFACTS parameter

mike-scott:
Add support for EXTRA_ARTIFACTS parameter.  This parameter would be
used to add a "special" (non-standard) build output into the
"Artifacts" list.

### DIFF
--- a/lmp/build.sh
+++ b/lmp/build.sh
@@ -130,6 +130,13 @@ if [ -d "${archive}" ] ; then
 	cp ${DEPLOY_DIR_IMAGE}/bbl* ${archive}/other/ || true
 	cp ${DEPLOY_DIR_IMAGE}/fw_* ${archive}/other/ || true
 
+	# Handle user provided extra artifacts
+	if [ ! -z "${EXTRA_ARTIFACTS}" ]; then
+		for extra_file in ${EXTRA_ARTIFACTS}; do
+			cp ${DEPLOY_DIR_IMAGE}/${extra_file} ${archive}/
+		done
+	fi
+
 	# Make the main img.gz be in the root of the archive
 	mv ${archive}/other/lmp-*.img.gz ${archive}/ || true
 	mv ${archive}/other/boot*.img ${archive}/ || true


mike-scott:
Re-Running this job:
https://ci.foundries.io/projects/aeler/lmp/builds/27/build-aeler-imx7ulpea-ucom-mfgtools/

Using this version of ci-scripts and an added parameter:
"EXTRA_ARTIFACTS": "mfgtools.tar.gz",

mike-scott:
Works as good as it looks:
`simulate [{'file': 'mfgtools.tar.gz', 'size': 12896655},`

doanac:
> Works as good as it looks:
I'm saving that exact same comment for the next time you send a bad PR :)

One thing we should go ahead and add here is a check in bb-config.sh like: https://github.com/foundriesio/ci-scripts/blob/master/lmp/bb-config.sh#L36 to enable this

lmp/build.sh
```diff
@@ -130,6 +130,13 @@ if [ -d "${archive}" ] ; then
 	cp ${DEPLOY_DIR_IMAGE}/bbl* ${archive}/other/ || true
 	cp ${DEPLOY_DIR_IMAGE}/fw_* ${archive}/other/ || true
 
+	# Handle user provided extra artifacts
+	if [ ! -z "${EXTRA_ARTIFACTS}" ]; then
+		for extra_file in ${EXTRA_ARTIFACTS}; do
+			cp ${DEPLOY_DIR_IMAGE}/${extra_file} ${archive}/
```
ricardosalveti: Should we publish these additional files to ${archive}/other as done with the files that are not the final image?

mike-scott: In this case the user has explicitly asked for the artifact and I was placing them at the top level for visibility.

mike-scott:
> > Works as good as it looks:
> 
> I'm saving that exact same comment for the next time you send a bad PR :)
> 
> One thing we should go ahead and add here is a check in bb-config.sh like: https://github.com/foundriesio/ci-scripts/blob/master/lmp/bb-config.sh#L36 to enable this

Do we need this in the local.conf?  It's post-OE processing step.

PR #29, opened 2019-10-01
factory-containers: add caching speedups, and modification detection.

EmbeddedAndroid:
This should give us a few improvements:

- use the --cache-from build flag to speedup rebuild
- detect when container files have changed, then trigger a build
- if the build has not changed, retag the images so that the dockerapps can be published

### DIFF
--- a/factory-containers/build.sh
+++ b/factory-containers/build.sh
@@ -84,24 +84,23 @@ for x in $IMAGES ; do
 	if [ -f /secrets/osftok ] ; then
 		status "Doing docker-login to hub.foundries.io with secret"
 		docker login hub.foundries.io --username=doesntmatter --password=$(cat /secrets/osftok) | indent
-		# sanity check and pull in a cached image if it exists
-		run docker pull ${ct_base}:${GIT_OLD_SHA:0:7}-$ARCH || echo "WARNING - no cached image"
+		# sanity check and pull in a cached image if it exists. if it can't be pulled set no_op_tag to 0.
+		run docker pull ${ct_base}:${GIT_OLD_SHA:0:7}-$ARCH || no_op_tag=0
+		if [ $no_op_tag -eq 0 ] && [ -z "$CHANGED" ]; then
+			echo "WARNING - no cached image found, forcing a rebuild"
+		fi
 		auth=1
 	fi
 
 	cd $x
-<<<<<<< HEAD
-	run docker build --cache-from ${ct_base}:${GIT_OLD_SHA:0:7}-$ARCH -t ${ct_base}:$TAG-$ARCH --force-rm .
-=======
 	if [ $no_op_tag -eq 1 ] ; then
 		status Tagging docker image $x for $ARCH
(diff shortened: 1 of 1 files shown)


EmbeddedAndroid:
I need to rebase this, shouldn't have ad the ota_test patch in this series

EmbeddedAndroid:
@doanac ok this is working as expected now in my factory.

factory-containers/build.sh
```diff
@@ -84,9 +89,19 @@ for x in $IMAGES ; do
 		auth=1
 	fi
 
-	status Building docker image $x for $ARCH
 	cd $x
+<<<<<<< HEAD
```
EmbeddedAndroid: When I fixed these merge conflicts, I ended up committing them to the next patch. Let me know if you want me to fix this here.

doanac:
looks cool. I guess there's no reason you'd ever want to override this, right? ie - its basically got all the logic to always do the "right thing", correct?

doanac:
ie - If there's a vulnerability in the containers "FROM", this change is always going to pull in that change, right?

EmbeddedAndroid:
> looks cool. I guess there's no reason you'd ever want to override this, right? ie - its basically got all the logic to always do the "right thing", correct?

Thinking about this again, I think we will want a way to force rebuild all containers. Let me add back in the CLEAN_BUILD env variable.

EmbeddedAndroid:
> ie - If there's a vulnerability in the containers "FROM", this change is always going to pull in that change, right?

It seems to update the base image no matter what see below:

```
== 2019-10-01 21:57:24 Running: docker build --cache-from hub.foundries.io/magicman/opencpn:526eee8-amd64 -t hub.foundries.io/magicman/opencpn:9782bff-amd64 --force-rm .
|  Sending build context to Docker daemon  3.584kB

|  Step 1/6 : FROM ubuntu:latest
|  latest: Pulling from library/ubuntu
|  5667fdb72017: Already exists
|  d83811f270d5: Already exists
|  ee671aafb583: Already exists
|  7fc152dfb3a6: Already exists
|  Digest: sha256:b88f8848e9a1a4e4558ba7cfc4acc5879e1d0e7ac06401409062ad2627e6fb58
|  Status: Downloaded newer image for ubuntu:latest
|   ---> 2ca708c1c9cc
|  Step 2/6 : RUN apt-get update  && apt-get install -y software-properties-common  && add-apt-repository ppa:opencpn/opencpn  && apt-get update  && apt-get install -y opencpn sudo
|   ---> Using cache
|   ---> be48f72191cf
```
It however uses cache from the subsequent layers which is the right behavior I believe.

doanac:
I think this all looks great then. We should probably add a comment into this script that explains these different scenarios and how the script always winds up doing the right thing.

EmbeddedAndroid:
@doanac ok I added a comment block at the top of the script, listing what scenarios should be supported. I also tested the NOCACHE flag in my factory and it's working as expected. If you are good with these, +1 to merge.

PR #37, opened 2019-11-20
factory-containers: drop OLD_TAG for latest reference

EmbeddedAndroid:
To avoid the case where the OLD_TAG image wasn't built, use the latest tag for cache instead. This should avoid use having to rebuild an image without cache if something happened to the previous build.


EmbeddedAndroid:
Been testing this change in my factory, good to go.

PR #39, opened 2019-11-25
factory-containers: Allow custom Dockerfiles and Build Contexts

doanac:


factory-containers/build.sh
```diff
@@ -68,7 +68,7 @@ for x in $IMAGES ; do
 			status "Detected changes to $x"
 		else
 			status "No changes to $x, tagging only"
-			no_op_tag=1
+			# no_op_tag=1
```
EmbeddedAndroid: Is this a typo?

mike-scott: Heh, I was just going to ask the same question

doanac: great catch. I commented that out to force some local debugging to work.

factory-containers/build.sh
```diff
@@ -100,23 +101,32 @@ for x in $IMAGES ; do
...
 			status Building docker image $x for $ARCH with cache
-			run docker build --label "jobserv_build=$H_BUILD" --cache-from ${ct_base}:latest -t ${ct_base}:$TAG-$ARCH --force-rm .
+			docker_cmd="$docker_cmd  --cache-from ${ct_base}:latest"
 		else
 			status Building docker image $x for $ARCH with no cache
-			run docker build --label "jobserv_build=$H_BUILD" --no-cache -t ${ct_base}:$TAG-$ARCH --force-rm .
+			docker_cmd="$docker_cmd  --no-cache"
 		fi
+
+		DOCKERFILE="${DOCKERFILE-Dockerfile}"
```
gwendolynbrook: Might need to move this up -- need `-f DOCKERFILE` in docker_cmd, I think.

factory-containers/build.sh
```diff
@@ -100,23 +101,32 @@ for x in $IMAGES ; do
...
+		if [ -n "$BUILD_CONTEXT" ] ; then
+			status "Using custom build context $BUILD_CONTEXT"
+			BUILD_CONTEXT="$REPO_ROOT/$x/${BUILD_CONTEXT}"
```
gwendolynbrook: hrm -- build context is relative to the $x/ directory, as opposed to the $REPO_ROOT?

In my current example, I'd need to set `BUILD_CONTEXT=../` -- is that an intended constraint?

factory-containers/build.sh
doanac: Ugh - my code only defines it. Its not even using it. I'll fix.

factory-containers/build.sh
doanac: honestly - i was confused which directory it should be relative to. My thinking was to make it relative to the directory the docker-build.conf file lived in. But yes, I think `BUILD_CONTEXT=../` shoud "just work".

factory-containers/build.sh
doanac: okay - i've force-pushed a fix to this branch. Its treating Dockefile relative to $REPO_ROOT/$x like I did for BUILD_CONTEXT so that things are consitent.

PR #43, opened 2019-12-10
lmp/jobserv.yml: add job for build release stable

ricardosalveti:
Similar to build-release, but following the lmp-manifest stable branch
(currently zeus). AKLITE_TAG and OTA_LITE_TAG also extended with -stable.

### DIFF
--- a/lmp/jobserv.yml
+++ b/lmp/jobserv.yml
@@ -88,6 +88,75 @@ triggers:
         persistent-volumes:
           bitbake: /var/cache/bitbake
 
+  - name: build-release-stable
+    type: git_poller
+    email:
+      users: 'ci-notifications@foundries.io'
+    params:
+      GIT_URL: |
+        https://github.com/foundriesio/lmp-manifest.git
+      GIT_POLL_REFS: |
+        refs/heads/zeus
+      OTA_LITE_TAG: postmerge-stable
+      AKLITE_TAG: promoted-stable
+    runs:
+      # supported images
+      - name: supported-{loop}
+        container: hub.foundries.io/lmp-sdk
+        host-tag: amd64-osf
+        loop-on:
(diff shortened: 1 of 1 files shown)


lmp/jobserv.yml
```diff
@@ -88,6 +88,95 @@ triggers:
...
+    type: git_poller
+    email:
+      users: 'ci-notifications@foundries.io'
+    params:
+      GIT_URL: |
+        https://github.com/foundriesio/lmp-manifest.git
+      GIT_POLL_REFS: |
+        refs/heads/zeus
+      OTA_LITE_TAG: postmerge-stable
+      AKLITE_TAG: promoted-stable
```
doanac: ugh - forgot about the confusion on that. Should we call this tag "zeus" instead or is "stable" going to make the most sense?

lmp/jobserv.yml
```diff
@@ -88,6 +88,95 @@ triggers:
...
+      # supported images that have OTA
+      - name: supported-{loop}
+        container: hub.foundries.io/lmp-sdk
+        host-tag: amd64-osf
+        loop-on:
+          - param: MACHINE
+            values:
+              - raspberrypi3-64
+        triggers:
+          - name: ota-{loop}
```
doanac: Do we want to trigger OTAs now or hold off on that?

lmp/jobserv.yml
ricardosalveti: I was thinking stable because we will be based on different OE versions as we go, but we will only support one stable stream (which will be based on the latest OE stable).

lmp/jobserv.yml
ricardosalveti: Guess better to hold off for now, let me remove this piece.

PR #49, opened 2020-01-15
factory-containers/build: cache: fall back to using git sha tag 

EmbeddedAndroid:
When you create a new OTA lite tag, via branch or git tag, a cache image with the ota lite tag will not exist. In this case, try to pull the git sha tag from a previous build if possible and use that to cache the build.


EmbeddedAndroid:
I completely agree, it's getting way too complicated.

PR #50, opened 2020-01-21
Allow containers to use secrets defined in JobServ

doanac:
JobServ puts all project defined secrets as files under /secrets. This change allows a user to *securely* pass these secrets into a docker build context so that they may be used to do things like grab binaries from a secure registry.

### DIFF
--- a/factory-containers/build.sh
+++ b/factory-containers/build.sh
@@ -18,18 +18,15 @@ require_params FACTORY
 
 run apk --no-cache add file git
 
+# required for "docker manifest"
+export  DOCKER_CLI_EXPERIMENTAL=enabled
+
 MANIFEST_PLATFORMS_DEFAULT="${MANIFEST_PLATFORMS_DEFAULT-linux/amd64,linux/arm,linux/arm64}"
 status Default container platforms will be: $MANIFEST_PLATFORMS_DEFAULT
 
 ARCH=amd64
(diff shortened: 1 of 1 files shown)


factory-containers/build.sh
```diff
@@ -144,14 +141,17 @@ for x in $IMAGES ; do
...
+		#  linux/amd64,linux/arm,linux/arm64
+		# into amd64 arm arm64
+		manifest_args=""
+		for arch in `echo $MANIFEST_PLATFORMS | sed -e 's/linux\///g' -e 's/,/ /g'` ; do
+			manifest_args="${manifest_args} ${ct_base}:$TAG-$arch"
+		done
+		run docker manifest create ${ct_base}:$TAG $manifest_args && \
+			run docker manifest create ${ct_base}:$LATEST $manifest_args && \
+			run docker manifest push ${ct_base}:$TAG && \
+			run docker manifest push ${ct_base}:$LATEST || true
```
EmbeddedAndroid: Finally, we can stop using that random github tool :)

doanac:
@EmbeddedAndroid - please pull in and see if it works for you.

EmbeddedAndroid:
@doanac lgtm, I've pushed a few builds through and had no problems with the new docker manifest / secrets code. +1 to merge.

PR #55, opened 2020-03-04
containers: Add a more reliable caching strategy

doanac:
This change allows us to find the most relevant previous container that
we can use as a build cache.


doanac:
@EmbeddedAndroid - this is totally untested. I think there's probably a bug in the case that it kind find the previous cache. Something that build.sh would need to be smarter about.

doanac:
This was merged in another pr.

PR #56, opened 2020-03-13
Include source shas in TUF Target

doanac:
This change helps track the code source SHA's that produce a given
target. This will be useful for tools like fioctl, the webui, and the
lmp platform itself.

### DIFF
--- a/lmp/customize-target
+++ b/lmp/customize-target
@@ -7,6 +7,7 @@ import argparse
 import logging
 import json
 import os
+import subprocess
 
 from copy import deepcopy
 
@@ -34,6 +35,12 @@ class TagMgr:
         return self._tags
 
 
+def git_hash(gitdir):
+    return subprocess.check_output(
+        ['git', 'log', '-1', '--format=%H'], cwd=gitdir
+    ).strip().decode()
+
+
 def merge(args):
     with open(args.targets_json) as f:
         data = json.load(f)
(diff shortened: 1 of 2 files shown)


doanac:
This produces a targets.json entry like:
~~~
"raspberrypi3-64-lmp-9" : {
            "custom" : {
               ...
               "containers-sha" : "e1947ccfa0debf9dde9138f273d1c67cf22ff551",             
               "lmp-manifest-sha" : "629e974bf4fd5fe7eb6c15bd99e52cb1bed9d372",
               "meta-subscriber-overrides-sha" : "d56ae6a677316bf1c8544cf9228632a59fe3d991",
               "name" : "raspberrypi3-64-lmp",
               "tags" : [
                  "master"
               ],
         },
~~~

doanac:
with this change we could update aktualizr lite to dump this information in /var/sota/current-target in a way that should solve our user's request.

lmp/customize-target
```diff
@@ -48,6 +55,9 @@ def merge(args):
     updates = []
     for idx, (tgt_tag, apps_tag) in enumerate(tagmgr.tags):
         tgt = targets[args.target_name]
+        tgt['custom']['lmp-manifest-sha'] = git_hash('/srv/oe/.repo/manifests')
+        tgt['custom']['meta-subscriber-overrides-sha'] = git_hash(
+            '/srv/oe/layers/meta-subscriber-overrides')
```
ricardosalveti: is is also going to be called for the public builds or just for subscribers?

doanac: good catch! force push fixed

PR #57, opened 2020-03-24
lmp: Support installed_versions in customize-target

doanac:


doanac:
This relies on the LMP having something like: https://github.com/foundriesio/ci-scripts/pull/57

PR #59, opened 2020-05-21
lmp/bb-config: add support to build ptest lmp variants via tags

ricardosalveti:
Allow building a special lmp build variant which includes the ptest
packages by default, but via a special tag in order to avoid
contaminating the base builds.

This is done by simply changing the OTA_LITE_TAG variable and appending
-ptest to the lmp-based tags, which is later consumed by the build and
published under a different tag than the one originally intended.

Using a different tag for the ptest-enabled builds allows the user to
easily switch to the ptest tag when performing the OE package testing,
and move back to the base image once that is concluded.

### DIFF
--- a/lmp/bb-config.sh
+++ b/lmp/bb-config.sh
@@ -11,6 +11,7 @@ H_BUILD="${H_BUILD-lmp-localdev}"
 LMP_VERSION=$(git --git-dir=.repo/manifests/.git describe --tags 2>/dev/null || echo unknown)
 FACTORY="${FACTORY-lmp}"
 UBOOT_SIGN_ENABLE="${UBOOT_SIGN_ENABLE-0}"
+ENABLE_PTEST="${ENABLE_PTEST-0}"
 
 source setup-environment build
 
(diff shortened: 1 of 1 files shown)


lmp/bb-config.sh
```diff
@@ -47,6 +48,30 @@ IMAGE_INSTALL_append = " ${EXTRA_IMAGE_INSTALL}"
...
 if [ -n "$OTA_LITE_TAG" ] ; then
+	# Ptest-based builds require the same build settings and variables,
+	# but the final image needs to be tagged differently, such as
+	# <main tag>-ptest, so perform the change at the OTA_LITE_TAG variable
+	if [ "${ENABLE_PTEST}" = "1" ]; then
+		IFS=","
+		PTAGS=""
+		for tag in ${OTA_LITE_TAG}; do
+			lmptag=$(echo $tag | cut -d: -f1)
+			PTAGS="${PTAGS} ${lmptag}-ptest"
```
doanac, line 59: It took me a minute to sort out how clever using a "space" was here instead of a comma.

PR #61, opened 2020-05-28
Disable docker build with file

gwendolynbrook:
Modify main loop build script to disable docker builds for directories that aren't meant to be images.

### DIFF
--- a/factory-containers/build.sh
+++ b/factory-containers/build.sh
@@ -57,8 +57,13 @@ trap 'echo "</testsuite>" >> /archive/junit.xml' TERM INT EXIT
 
 REPO_ROOT=$(pwd)
 for x in $IMAGES ; do
-	# Skip building things that end with .disabled
-	echo $x | grep -q -E \\.disabled$ && continue
+  # Skip building directories that end with .disabled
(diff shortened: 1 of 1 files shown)


mike-scott:
@gwendolynbrook We should have @doanac take a look, but we probably want to squash the 2 commits into 1.

From a functionality point of view this looks good tho.

doanac:
I've been toying with something similar that we are about to merge into master:

 https://github.com/foundriesio/ci-scripts/commit/c1815b7544a8536dd9f756173df0dba0c107b05e

This change is doing:
~~~
IMAGES=$(find ./ -mindepth 2 -maxdepth 2 -name Dockerfile | cut -d / -f2)
~~~

Would that work for you? Or are you wanting to put .disable on directories that have a Dockerfile?

gwendolynbrook:
> IMAGES=$(find ./ -mindepth 2 -maxdepth 2 -name Dockerfile | cut -d / -f2)

@doanac this would cover my current need; I don't presently have a case where there's a directory I want to disable that *does* have a dockerfile.

doanac:
I've just merge: https://github.com/foundriesio/ci-scripts/pull/62

PR #65, opened 2020-06-02
containers: Fix for the compose-app only case

mike-sul:


mike-sul:
It works in all use-cases except the case when it becomes zero/no apps in 'containers'. A new target includes all apps that the previous target includes.

PR #70, opened 2020-07-08
containers: Print out progress

doanac:
This will allow jobserv to print the %complete while a run is active

### DIFF
--- a/factory-containers/build.sh
+++ b/factory-containers/build.sh
@@ -56,8 +56,13 @@ fi
 echo '<testsuite name="unit-tests">' > /archive/junit.xml
 trap 'echo "</testsuite>" >> /archive/junit.xml' TERM INT EXIT
 
+total=$(echo $IMAGES | wc -w)
+total=$((total*3)) # 3 steps per container: build, push, test*manifest
+completed=-3  # we increment on the first step of the first loop.
+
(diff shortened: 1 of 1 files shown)


doanac:
Here's what it looks like afterwards when you have it in your factory:

 https://ci.foundries.io/projects/andy-corp/lmp/builds/98/build-amd64-partner/artifacts/console.log

factory-containers/build.sh
```diff
@@ -151,6 +158,8 @@ for x in $IMAGES ; do
 			run docker push ${ct_base}:$TAG-$t
 		done
 
+		echo "Build step $((completed+2)) of $total is complete"
+
```
mike-scott, line 162: Seems like we could simplify and combine this line and line 176 down under the `fi` on line 177?

I know technically we're also doing a docker manifest create in this block, but the customer won't really know that and it's almost like that's being included in step 2.

doanac: yep. and with the webui checking every 3 seconds, its unlikely this step will even get picked up for showing status.

doanac: actually - we could remove the 2nd instance of this. its okay that steps are "skipped". and then for 99% of the time we get slightly more granular status. I'll do that in my force-push update.

PR #71, opened 2020-07-16
compose app images dumping and preloading

mike-sul:
- dump all container images of each app into a tar file, format
dockerd:overlay2;
- load the dumped images into var/lib/docker during LmP build.


factory-containers/publish-compose-apps
```diff
@@ -135,22 +139,55 @@ def publish(factory: str, tag: str, app_name: str) -> str:
...
+
+        status('Dumping container images of the published compose apps to ' + app_images_dir)
+
+        prefixed_platforms = platforms.split(',')
+        archs = []
+        for platform in prefixed_platforms:
+            archs.append(platform.split('/')[1])
+
+        with tempfile.TemporaryDirectory() as tmp_dst_dir:
+            dump_app_images(os.path.abspath(os.getcwd()), archs, tmp_dst_dir)
```
doanac, line 179: i think you can replace `os.path.abspath(os.getcwd())` with the `cur_work_dir` you've added on line 146

factory-containers/publish-compose-apps
```diff
@@ -135,22 +139,55 @@ def publish(factory: str, tag: str, app_name: str) -> str:
     return app + '@sha256:' + sha.decode()
 
 
-def main(factory: str, tag: str):
+def main(factory: str, tag: str, platforms: str, sha: str, app_preload_flag: str, app_dir=None):
```
doanac, line 142: trying to understand the addition of `app_dir`. does this make it easier for you to do unit/manual testing?

lmp/preload-app-images
```diff
@@ -0,0 +1,48 @@
...
+
+
+def preload_app_images(target_json_file_filepath, dst_dir, arch):
+    with open(target_json_file_filepath) as target_json_file:
+        target_json = json.load(target_json_file)
+
+    target_name = list(target_json.keys())[0]
+    logging.info('Preloading app images for ' + target_name)
+    sha = target_json[target_name]['custom']['containers-sha']
+    image_data_dir = os.path.join('/var/cache/bitbake/app-images/', sha)
```
doanac, line 31: I guess this directory will grow over time. I've added a note in Jira about this so we remember to attempt to prune old data eventually: https://foundriesio.atlassian.net/jira/software/projects/FF/boards/13?selectedIssue=FF-11

doanac: we might be able to reason about what to delete by looking at targets.json and understanding that no new LMP builds would use an existing tarball, so it could be safe to delete. then we'd basically always keep the minimal amout of tarballs around. but that can be added further down the road

factory-containers/compose_app_downloader.py
```diff
@@ -0,0 +1,217 @@
...
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self._process.terminate()
+        self._process.wait(timeout=60)
+
+        logger.info("Containerd daemon has been stopped")
+        shutil.rmtree(self._container_dir)
+
+
+class DockerDaemon:
```
doanac, line 63: we already have dockerd running when this code gets hit. Do you need your own instance to make sure you have a pristine environment?

factory-containers/compose_app_downloader.py
```diff
@@ -0,0 +1,217 @@
...
+
+class ImageDownloader:
+    def __init__(self, env):
+        self._env = os.environ.copy()
+        self._env.update(env)
+
+    def pull(self, platform, url):
+        cmd = self._get_cmd(platform, url)
+        logger.info('Fetching image: {}'.format(cmd))
+        subprocess.run(cmd.split(), check=True, stdout=None, stderr=None, env=self._env)
```
doanac, line 114: think you need to check the return value of subprocess.run or do a subprocess.check_call which will raise an error if it fails

factory-containers/compose_app_downloader.py
```diff
@@ -0,0 +1,217 @@
...
+    with ContainerDaemon() as containerd:
+        with DockerDaemon(dst_dir, containerd.address) as dockerd:
+            downloader = image_downloader_cls(dockerd.host)
+
+            def download_app(app: ComposeApp):
+                for service_name, service_cfg in app.services():
+                    logger.info('Copying image {} to {}'.format(service_cfg['image'], dst_dir))
+                    downloader.pull(arch, service_cfg['image'])
+
+            yield download_app
```
doanac, line 168: slick use of chaining together some context managers.

factory-containers/compose_app_downloader.py
mike-sul: Yes, there are a few reasons for that (the one you mentioned is valid too):
1. a host environment executing the given may have or may not have dockerd running. The given approach makes the implementation less vulnerable/sensitive to a host env, so it can work on a local host, VM or any other container as long as dockerd is installed;

2. The host dockerd's data root (/var/lib/docker/) can be polluted by data of some other images while we need data/files that belongs solely to the target app images;

3. Using `--platform` option requires running dockerd in the experimental mode. `"--platform" is only supported on a Docker daemon with experimental features enabled`. This option is required because we want to dump images of other than a host architecture (dumping arm64's images on amd64 host).

4. It allows dumpling images of all target architectures concurrently, multiple instances of DockerDaemon can be running concurrently without interfering each other what can speed up an overall process. Not implemented yet though :)

factory-containers/compose_app_downloader.py
mike-sul: I think `check=True` param makes it raise an exception if the command/process fails. Anyway, I see that I need to improve the implementation in terms of exception handling, I think I should catch exception within the given module. Thanks.

factory-containers/publish-compose-apps
mike-sul: It makes it easier to do manual testing. I can remove it.

factory-containers/publish-compose-apps
mike-sul: line #148 can change a cur dir, so we have to get it once more here. cur_work_dir is needed here to make sure that execution environment comes back to the original cur work dir. But, all that stuff can be removed, I basically need it for unit & manual testing, it allows me running this script from ci-scripts directory and points it to my factory's containers dir.

lmp/preload-app-images
mike-sul: I consider it as a temporal, putting image data is simple and straightforward and good for a start, but it has many drawbacks.

- First of all, it requires the container and lmp builds to have access to the same NFS volume;
- secondary, this storage has size limits and is not redundant and so reliable as GCS or S3; 
- Once we move towards using ostree as a storage for container images we will have to store images somewhere on the cloud (most likely Treehub backed by S3 for the beginning) so devices have access to it.

factory-containers/compose_app_downloader.py
mike-sul: Added (forced push) exception handling in the main.

factory-containers/publish-compose-apps
doanac: no. i just wanted to make sure i was reading things correctly. it seems useful.

factory-containers/publish-compose-apps
doanac: this all kind of makes sense. the code might be a little less complex if this test-env specific logic was moved into the the __main__ section, but I'm not sure. this is tested, so i'd say leave it.

factory-containers/compose_app_downloader.py
mike-sul: @doanac https://github.com/foundriesio/meta-lmp/pull/195 has been merged, so I suppose we should merge this PR too.

ricardosalveti:
Please add a license/copyright to the new files (not every file got one, something we should fix, but we can get the new ones right).

e.g.:
```
# Copyright (c) 2020 Foundries.io
# SPDX-License-Identifier: Apache-2.0
```

factory-containers/compose_app_downloader.py
```diff
@@ -0,0 +1,220 @@
...
+        shutil.rmtree(self._container_dir)
+
+
+class DockerDaemon:
+    ImageDataDirs = ['overlay2', 'image']
+
+    def __init__(self, dst_dir: str, containerd_addr: str):
+        self.data_root = dst_dir
+        self._containerd_addr = containerd_addr
+        #self.cmd = '/home/mike/bin/dockerd-rootless.sh'
```
ricardosalveti, line 69: Please remove this line if not really required.

ricardosalveti:
Just some minor comments, but LGTM, will let @doanac do the final +1 here.

factory-containers/compose_app_downloader.py
mike-sul: > Just some minor comments, but LGTM, will let @doanac do the final +1 here.

Updated (forced push) according to the comments.

lmp/preload-app-images
```diff
@@ -0,0 +1,52 @@
...
+
+    arch_map = {'aarch64': 'arm64', 'x86_64': 'amd64', 'arm': 'arm'}
+    image_data_tar = os.path.join(image_data_dir, '{}-{}.tar'.format(sha[:7], arch_map[arch]))
```
doanac, line 38: just noticed this - `sha[:7]` isn't actually always the case:
  https://github.com/foundriesio/ci-scripts/commit/8edbdad88b882d91bd4b1316c7197c071ef8d0fd

mike-sul: Are you implying to use a full hash/sha value? BTW, we use it in other places, e.g. https://github.com/foundriesio/ci-scripts/blob/321d877d80f560b24ef99f221685e1b2a341f231/factory-containers/publish-compose-apps#L112 (tag is 7 symbol hash).

doanac: not the full hash but tag needs to be computed correctly. In the example you shared its actually getting that from the TAG environment variable set in docker-app-publish.sh which uses `git log -1 --format=%h`. Basically what I'm saying is 99% of the time the short-hash (TAG) is 7 characters. However, we occasionally have exceptions, so you have to use `git log -1 --format=%h` to ensure you get the correct tag.

mike-sul: understood. In this case, I should think of something else as the given script is called during lmp build while 'sha' is a sha of corresponding containers repo commit so running `git log -1 --format=%h` will yield an abbreviated hash of the current lmp commit.

mike-sul: updated

PR #74, opened 2020-08-07
containers: Specify Targets to assemble a system image for

mike-sul:

### DIFF
--- a/factory-containers/assemble-system-image
+++ b/factory-containers/assemble-system-image
@@ -130,6 +130,9 @@ def get_args():
     parser.add_argument('-o', '--out-image-dir', help='A path to directory to put a resultant image to',
                         default=os.environ.get('OUT_IMAGE_DIR'))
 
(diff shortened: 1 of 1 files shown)


factory-containers/assemble-system-image
```diff
@@ -181,7 +184,11 @@ if __name__ == '__main__':
     wic_tool = args.wic_tool
 
     try:
-        targets = get_targets_from_api(args.factory, args.build_num, args.token)
+        if args.targets:
+            targets = json.loads(args.targets)
```
doanac, line 188: so this is a list of full blown Tuf Targets, right? ie - its not just a list target keys? Is it possible to share a little more context on how this is going to get called?

mike-sul: >so this is a list of full blown Tuf Targets, right?
yes

The JobServe CI job that runs this script will be triggered by the ota-lite service, the ota-lite part will be triggered by fioctl as we discussed. The ota-lite API/handler will be HTTP POST targets/<target>/images.

PR #75, opened 2020-08-12
containers: Dump container images if not found

mike-sul:
If container images of Compose Apps are not found on
a local NFS/share/folder then try to download/dump them
during assembling of a system image with "pre-loaded" conatiners.

### DIFF
--- a/factory-containers/assemble-system-image
+++ b/factory-containers/assemble-system-image
@@ -5,36 +5,51 @@
 #
 import subprocess
 import os
-import requests
 import json
 import argparse
 import logging
+import tempfile
+import base64
+
+from helpers import http_get
+from compose_app_downloader import dump_app_images
+
 
 logger = logging.getLogger("System Image Assembler")
 wic_tool = None
 
 
-def get_targets_from_api(factory: str, build_num: str, token: str):
-    logger.info('Fetching factory targets; factory: {}, version: {}'.format(factory, build_num))
-
+def _get_targets_from_api(factory: str, token: str):
     api_base_url = 'https://api.foundries.io/ota/repo/'
     targets_endpoint = 'api/v1/user_repo/targets.json'
 
-    res_targets = {}
     target_url = os.path.join(api_base_url, factory, targets_endpoint)
-    target_resp = requests.get(target_url, headers={'OSF-TOKEN': token})
-    if not target_resp.ok:
-        raise requests.exceptions.HTTPError('Failed to get {}: HTTP_{}\n{}'.
-                                            format(target_url, target_resp.status_code, target_resp.text))
+    target_resp = http_get(target_url, headers={'OSF-TOKEN': token})
+    return target_resp.json()['signed']['targets']
 
-    for target_name, target in target_resp.json()['signed']['targets'].items():
+
+def get_targets_from_api_by_build_numb(factory: str, build_num: str, token: str):
+    logger.info('Fetching factory targets; Factory: {}, version: {}'.format(factory, build_num))
+    targets = _get_targets_from_api(factory, token)
+    res_targets = {}
+
+    for target_name, target in targets.items():
         custom = target.get('custom')
         if not custom:
             continue
 
         if custom.get('version', '') == build_num and custom.get('targetFormat', 'NONE') == 'OSTREE':
             res_targets[target_name] = target
+    return res_targets
 
+
+def get_targets_from_api_by_targets(factory: str, target_names: list, token: str):
+    logger.info('Fetching factory targets; Factory: {}, Targets: {}'.format(factory, target_names))
+    targets = _get_targets_from_api(factory, token)
+    res_targets = {}
+    for target_name in target_names:
+        if target_name in targets:
+            res_targets[target_name] = targets[target_name]
     return res_targets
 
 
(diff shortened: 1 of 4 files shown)


factory-containers/assemble-system-image
```diff
@@ -60,7 +81,86 @@ def get_system_image_from_ci(target: dict, token: str):
...
+        'Authorization': 'Basic ' + base64.b64encode(user_pass.encode()).decode()
+    }
+
+    params = {
+        'service': 'registry',
+        'scope': 'repository:{}/{}:pull'.format(factory, repo)
+    }
+
+    token_req = requests.get(registry_oauth_token, headers=headers, params=params)
+    token_req.raise_for_status()
```
doanac, line 97: might want to raise an HTTPException with the response text like you do up higher in the file. 
~~~
if not target_resp.ok:
        raise requests.exceptions.HTTPError('Failed to get {}: HTTP_{}\n{}'.
                                            format(target_url, target_resp.status_code, target_resp.text))
~~~

doanac: sometime this gets so tedious i write a single "requests_get" that does that and sets the osftok in the header.

doanac: in fact- i forgot - https://github.com/foundriesio/ci-scripts/blob/master/helpers.py#L112  - if that helps

factory-containers/assemble-system-image
```diff
@@ -69,6 +169,14 @@ def copy_container_images_to_wic(target: dict, app_image_dir: str, wic_image: st
     containers_arch = arch_map[target['custom']['arch']]
 
     app_image_tar_src = os.path.join(app_image_dir, containers_sha, '{}-{}.tar'.format(containers_sha, containers_arch))
+    if not os.path.exists(app_image_tar_src):
+        logger.info('Container images have not been found, trying to obtain them...')
+        os.makedirs(os.path.dirname(app_image_tar_src), exist_ok=True)
+        dump_apps_container_images(target, containers_arch, app_image_tar_src, token)
+
+    if not os.path.exists(app_image_tar_src):
```
doanac, line 177: can this condition happen? the if above for that same condition calls `makedirs`?

mike-sul: Yes, the above `makedirs` makes a dir while the given `if` verifies if the tar file exists, I am just not sure if in all failure cases exception will be generated so double-checking if the tar file containing container images actually exists.

factory-containers/assemble-system-image
mike-sul: it helps but not in this specific case since `jobserv_get` sets a specific header (osf-token) while in these three cases headers are different.

mike-sul:
@doanac updated according to your comments. I didn't use `jobserv_get` because it introduces implicit policy, i.e. a token must be in /secrets/osftok file. I would like the script to be free from such implicit requirement/policy and its users to specify a token location explicitly.

factory-containers/assemble-system-image
doanac: I think it should. If it doesn't the code will fail but should hopefully be debuggable. I'd do one of two things 1) remove this 2) put a comment that explains why.

mike-sul: The problem is that the following code won't fail if `app_image_tar_src` doesn't exist and it's really bad because it will look like the run/job is successful while the resultant wic image won't contain container images.
For some reason this call `subprocess.check_call([wic_tool, 'cp', app_image_tar_src, app_image_tar_dst])` is successful if `app_image_tar_src` does not exists. I'll double check it tomorrow.

mike-sul:
>I think it should. If it doesn't the code will fail but should hopefully be debuggable. I'd do one of two things 1) remove this 2) put a comment that explains why.

There is no guarantee that in the case of a `dump_apps_container_images` failure an exception or exit() will be called (some rough edge cases, we spawn a number 3rd party tools that don't always behave as expected).

Moreover, the following call `subprocess.check_call([wic_tool, 'cp', app_image_tar_src, app_image_tar_dst])` does not raise any exception if `app_image_tar_src` doesn't exist. Therefore, in order to detect such subtle/silent error the following check is enabled, it just throws an exception if the tar file does not exist prior to injecting it to the target wic file.

`bash-5.0# /usr/bin/oe/scripts/wic cp /archive/app-dir/5cd87f67a163d9dce788908c9f6d59a8066fd63c/5cd87f67a163d9dce788908c9f6d59a8066fd63c-arm64.tar lmp-factory-image-raspberrypi4-64.wic:2/ostree/deploy/lmp/var/sota/import/

bash-5.0# echo $?
0

bash-5.0# ls -al /archive/app-dir/5cd87f67a163d9dce788908c9f6d59a8066fd63c/5cd87f67a163d9dce788908c9f6d59a8066fd63c-arm64.tar
ls: /archive/app-dir/5cd87f67a163d9dce788908c9f6d59a8066fd63c/5cd87f67a163d9dce788908c9f6d59a8066fd63c-arm64.tar: No such file or directory
bash-5.0#`

factory-containers/assemble-system-image
doanac: Ugh - i read this wrong. the first you do a `mkdir(dirname(app_image_tar_src)` you then check if `app_image_tar_src` exists. that's two different things. sorry for the disctraction.

PR #79, opened 2020-08-27
containers: Resize wic before copying tarfile

doanac:
The wic image is created with a small overhead of free space. We should
increase the wic file's size to account for the amount it will grow by
when adding the tarfile into it.

### DIFF
--- a/factory-containers/assemble-system-image
+++ b/factory-containers/assemble-system-image
@@ -10,8 +10,9 @@ import argparse
 import logging
 import tempfile
 import base64
+from math import ceil
 
-from helpers import http_get
+from helpers import cmd, http_get
 from compose_app_downloader import dump_app_images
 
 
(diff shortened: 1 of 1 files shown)


factory-containers/assemble-system-image
```diff
@@ -150,6 +151,35 @@ def dump_apps_container_images(target: dict, containers_arch: str, app_image_tar
                                    '-C', os.path.join(tmp_docker_dir, containers_arch), '.'])
 
 
+def resize_wic(wic_image: str, increase_bytes: int):
+    increase_k = ceil(increase_bytes / 1024) + 1
+    wic_k = int(os.stat(wic_image).st_size / 1024)
```
mike-scott, line 156: wic_k will be rounded down, but increase_k will be rounded up... this *might* be ok?  Do we want to round both up?

doanac: i'll round up. i *think* that value is always a multiple of 1024, but your reasoning makes sense

factory-containers/assemble-system-image
```diff
@@ -150,6 +151,40 @@ def dump_apps_container_images(target: dict, containers_arch: str, app_image_tar
                                    '-C', os.path.join(tmp_docker_dir, containers_arch), '.'])
 
 
+def resize_wic(wic_image: str, increase_bytes: int):
+    increase_k = ceil(increase_bytes / 1024) + 1
+    wic_k = ceil(os.stat(wic_image).st_size / 1024)
+    count = wic_k + increase_k
+    cmd('dd', 'if=/dev/zero', 'bs=1024', 'of=' + wic_image,
+        'conv=notrunc', 'oflag=append', 'count=' + str(count),
+        'seek=' + str(wic_k))
```
mike-sul, line 160: if seek == wic_k then count should be increase_k?

factory-containers/assemble-system-image
```diff
@@ -178,9 +213,15 @@ def copy_container_images_to_wic(target: dict, app_image_dir: str, wic_image: st
     # Therefore, in order to detect such subtle/silent error the following check is enabled,
     # it just throws an exception if the tar file does not exist prior to injecting it to the target wic file.
 
-    if not os.path.exists(app_image_tar_src):
+    try:
+        st = os.stat(app_image_tar_src)
```
mike-sul, line 217: @mike-scott Do you think it's good enough for the juul case to increase the wic image enough to accommodate the tarball and on a first boot something (??) will increase rootfs further to accommodate the untarred image data? Or should we actually increase the wic image by the size of the untarred container image data?

factory-containers/assemble-system-image
doanac: that was how @mike-scott 's script worked. I think `count` and `seek` aren't related. ie - its going add count-seek bytes. right  mike?

mike-sul: -rw-r--r--  1 mike mike 911M Aug 28 21:04 17b406dc02163a515da0a05f8388c4ffea492cda-arm.tar
-rw-rw-r--  1 mike mike 699M Aug 28 10:56 lmp-factory-image-apalis-imx6.wic.org
-rw-rw-r--  1 mike mike 2.3G Aug 28 10:39 lmp-factory-image-apalis-imx6.wic

2.3 = 699 + (699 + 911)

mike-sul (approved):

PR #81, opened 2020-09-02
lmp: Support dynamically scoped token in creds.zip

doanac:
This change allows us to set the client_secret value of credentials.zip
during each build. This will allow us to remove this client_secret
value making the credentials.zip file a little more secure.

### DIFF
--- a/create-creds
+++ b/create-creds
@@ -1,0 +1,15 @@
+#!/usr/bin/python3
+import sys
+
+from helpers import generate_credential_tokens, require_secrets
+
+
+def main(creds_in, creds_out):
+    require_secrets('osftok', 'triggered-by')
+    generate_credential_tokens(creds_in, creds_out)
+
+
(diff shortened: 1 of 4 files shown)


doanac:
@mike-sul - you mind trying this out in your factory also? Just want to see it work for >1 credentails.zip files.

create-creds
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/python3
+import sys
+
+from helpers import generate_credential_tokens, require_secrets
+
+
+def main(creds_in, creds_out):
+    require_secrets('osftok', 'triggered-by')
+    generate_credential_tokens(creds_in, creds_out)
```
mike-sul, line 9: what's the lifetime of this token, just wondering if it won't expire during long LmP build.

mike-sul (approved):

PR #82, opened 2020-09-09
Source pre-build.conf before trying to find images

gwendolynbrook:
Run a pre-build script before any image build configuration.


doanac (approved):
thanks!

PR #83, opened 2020-09-14
lmp: Add update-targets script

doanac:
This was in the ota-lite repo and used to handle the `fioctl targets
tag` command. Its getting complex enough to live in ci-scripts where
we can take advantage of common code.

### DIFF
--- a/lmp/update-targets
+++ b/lmp/update-targets
@@ -1,0 +1,66 @@
+#!/bin/bash -e
+
+HERE=$(dirname $(readlink -f $0))
+source $HERE/../helpers.sh
+
+apk add curl
+
+CREDENTIALS=$(mktemp)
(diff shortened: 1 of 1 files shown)


mike-sul (approved):
AFAIU, the given script is exactly the same as it was in the ota-lite templates and has been proven over time.

doanac:
> AFAIU, the given script is exactly the same as it was in the ota-lite templates and has been proven over time.

Its almost the same. I changed some "echo" lines to use our "status" helper and I set up creds.zip properly (with the run's api token):
~~~
 CREDENTIALS=$(mktemp)
 $HERE/../create-creds /var/cache/bitbake/credentials.zip $CREDENTIALS
~~~

PR #86, opened 2020-09-29
containers: Remove Docker Apps support

mike-sul:
Signed-off-by: Mike Sul <mike.sul@foundries.io>

PR #87, opened 2020-09-30
Test webhook

mike-sul:


mike-sul:
Should go after https://github.com/foundriesio/ci-scripts/pull/86.
This is just an initial step/skeleton, going to add some more tests over time.

mike-sul:
@doanac What do you think about this initial step towards setting up some CI for this repo?

doanac (commented):
seems fine. but do we want both github action and jobserv in this? I think one or the other would be fine?

mike-sul:
> seems fine. but do we want both github action and jobserv in this? I think one or the other would be fine?

I thought about splitting the overall testing effort between them so we don't use JobServ's worker machines much as they used by our customers. Also, sometimes our worker pool so busy that it takes ages until some worker becomes available for the given tests. But, I don't have a strong opinion, can remove the github's actions.

The next step would be running some high-level/integration tests and it will require setting up some test factory as we discussed earlier.

doanac (approved):

PR #93, opened 2020-10-20
Refactor image preloading

mike-sul:
- Make ci-scripts testable;
- Refactor Apps' images preloading: testability, reusability, prep for applying ostree.


helpers.py
```diff
@@ -82,6 +82,7 @@ def cmd(*args, cwd=None, capture=False):
             out += line
     sys.stdout.buffer.write(b'|--\n')
     p.wait()
+    p.communicate()
```
doanac, line 85: interesting - i'm guessing you stumbled across some bug that required this?

apps/publish.py
```diff
@@ -0,0 +1,128 @@
...
+    apps.validate()
+    status('Compose Apps has been validated: {}'.format(apps.str))
+
+    status('Tagging Apps...')
+    allowed_tags = ['${TAG}', 'latest']
+    if factory == 'lmp':
+    # the lmp containers repo pulls in fiotest which sets
+                            # its tag to "postmerge" which is okay based on how
+                            # we do tagging for that repo.
+                            allowed_tags.append('postmerge')
```
doanac: strange indentation.

apps/target_manager.py
```diff
@@ -40,15 +49,18 @@ def create_target(targets_json, version, compose_apps, ota_lite_tag, git_sha):
...
                 del target['custom']['docker_apps']
-            logging.info('Targets with apps: %r', target)
+            logging.info('New Target with {}\n{}'.
+                         format(tgt_name, json.dumps(target, ensure_ascii=True, indent=2)))
+            new_target_counter += 1
 
     with open(targets_json, 'w') as f:
         json.dump(data, f, indent=2)
+
+    return new_target_counter > 0
```
doanac: can you articulate the changes here a bit. Some new things just look like changes to add more logging. But there are some other changes as well. I want to make sure I'm not missing something subtle

apps/compose_apps.py
```diff
@@ -0,0 +1,126 @@
...
+            self._run_cmd('config')
+
+        def services(self):
+            return self['services'].items()
+
+        def images(self):
+            for _, service_cfg in self.services():
+                yield service_cfg['image']
+
+        def tag_images(self, tag, image_base_url, allowed_tags=['${TAG}', 'latest']):
```
doanac: Think you want to default allowed_tags to None and then do an `if allowed_tags is None` in the actual method, or we'll find up with an unexpected python 'bug' some day.

doanac (commented):
nothing major on my first pass. I'll set aside some time to try this out for myself later this week.

helpers.py
mike-sul: Yeah, it's just hung in some cases from time to time. Also, it should be enough to have just p.communicate which also waits for the child process termination.

apps/publish.py
mike-sul: should be fixed.

apps/compose_apps.py
mike-sul: It's not relevant in the newest version that I force pushed.

apps/target_manager.py
mike-sul: New things:
1. The build number becomes optional. It's useful for testing purposes when you run it out of the scope of any CI/JobServ job. In this case, the build number (aka version) is calculated as the latest version found in the current targets.json + 1.

2. It returns a list of the new Targets that just been created. The follow-up "preloading" procedure needs a list of new Targets as an input so it knows what apps to dump/preload (apps' images arch/platform).

doanac:
this is starting to grow more. my biggest concern now is why the need for renaming the factory-containers directory. its not a backward compatible change. I don't think i can test this in my factory without hacking our git update hook. I don't understand the need to take this risk for what's a directory rename?

mike-sul:
> its not a backward compatible change. I don't think i can test this in my factory without hacking our git update hook.

The `factory-containers` is there https://github.com/mike-sul/ci-scripts/tree/refactor-image-preloading/factory-containers.
You can test it in your factory, no need for the git-update-hook hacking.

doanac:
works on my factory. I think the next step might be to update "repoz" to use this version for members of development

mike-sul:
> works on my factory. I think the next step might be to update "repoz" to use this version for members of development

do you mean something like this https://git.foundries.io/development/cloudplatforms/infrastructure/repoz/merge_requests/23/diffs ?

PR #95, opened 2020-11-03
container preload: Add some progress tracking

doanac:
Preload can take a while to run. This helps add some rough progress
tracking. We could take this and make a small jobserv change so that
these CI runs could start showing percent done in the UI.

### DIFF
--- a/assemble.py
+++ b/assemble.py
@@ -10,7 +10,7 @@ import argparse
 import logging
 from math import ceil
 
-from helpers import cmd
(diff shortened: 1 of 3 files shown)


doanac:
@mike-sul - feel free to just nak this and not even think about it. i found it handy while i was playing with this code over the weekend

mike-sul (approved):
It looks quite innocent (the key change is isolated) yet useful :+1:

PR #96, opened 2020-11-05
Consider MACHINES while creating Targets

mike-sul:
- Take into account currently configured MACHINEs for Factory while creating new Targets during the container build;
- Create new Targets only for those MACHINEs that have the latest Target version (some of the MACHINE builds could fail)
- Create new Targets only for those MACHINEs for which there is a compatible architecture of the containers being created.


mike-sul:
@doanac I verified all edge cases I planned to aim at the prod.

apps/target_manager.py
```diff
@@ -20,16 +21,25 @@ def create_target(targets_json, compose_apps, ota_lite_tag, git_sha, in_version=
...
+
         for name, target in data['targets'].items():
-            if target['custom']['targetFormat'] == 'OSTREE':
-                ver = int(target['custom']['version'])
-                latest_version = ver if latest_version < ver else latest_version
+            hwid = target['custom']['hardwareIds'][0]
+            ver = int(target['custom']['version'])
+            latest_version = ver if latest_version < ver else latest_version
+            # create new Targets based on existing Targets that
+            # 1. are OSTREE type (TODO: it's redundant since we have only ostree Targets)
```
doanac, line 31: are you sure we can really state its redundant? A user could choose to put non-ostree targets into this?

apps/target_manager.py
```diff
@@ -41,6 +51,33 @@ def create_target(targets_json, compose_apps, ota_lite_tag, git_sha, in_version=
...
+            # if it's not true for some parent Target then we skip creation of the new Target. It can be so
+            # in the case if the expected version of the parent Target failed and for some reason we don't
+            # want to create a new Target based on the latest available Target
+            if int(target['custom']['version']) != latest_tag_version[tag]:
+                logger.warning('Skipping Target creation: {}, corresponding latest parent Target has not been found'
+                               .format(tagmgr.create_target_name(target, version, tag),
+                                       tagmgr.create_target_name(target, str(latest_tag_version[tag]), tag)))
+                continue
+
+            # TODO: refactor it in order to bring to common denominator with the other parts of code
```
doanac, line 64: Can you put a comment on what you are checking for here? I'm having a hard time tracking this. Would splitting this up into some helper function make it any cleaner/clearer?

apps/target_manager.py
mike-sul: >A user could choose to put non-ostree targets into this?

How can a user create a non-ostree target? Anyway, it's not a big deal I can remove this comment.

apps/target_manager.py
mike-sul: >Can you put a comment on what you are checking for here?
Yes, I'll add a comment, effectively we discussed it last Thursday.

>  Would splitting this up into some helper function make it any cleaner/clearer?
Yes, makes sense, I'll try to make the code cleaner and more understandable.

apps/target_manager.py
mike-sul: I removed the TODO comment.

apps/target_manager.py
mike-sul: Restructured the code a little bit and added more comments along with the use cases' examples.

apps/publish.py
```diff
@@ -30,9 +30,11 @@ def main(factory: str, sha: str, targets_json: str, app_root_dir: str, publish_t
 
     apps_to_add_to_target = AppsPublisher(factory, publish_tool).publish(apps, apps_version)
 
-    status('Creating Target(s) that refers to the published Apps; tag: {}, version: {} '
-           .format(target_tag, target_version))
-    new_targets = create_target(targets_json, apps_to_add_to_target, target_tag, sha, target_version, new_targets_file)
+    status('Creating Targets that refer to the published Apps; tag: {}, version: {}, machines: {}, platfroms: {} '
```
doanac, line 33: `plafroms` -> `platforms`

apps/target_manager.py
```diff
@@ -41,6 +51,56 @@ def create_target(targets_json, compose_apps, ota_lite_tag, git_sha, in_version=
...
+            # 		hwids: x86_64
+            # 		tags: master
+            #
+            # factory_config.yml
+            # containers:
+            #   platforms:
+            #    - arm64
+            if not _can_apps_run_on_target_machine(target, platforms):
+                logger.warning('Skipping Target creation: {};'
+                               ' None of the container platforms is compatible with Target\'s architecture;'
```
doanac, line 98: s/is/are/

doanac (commented):
looks good. I just found two grammar nits. I usually don't like code that requires such extensive commenting, but in this case, i think it helps quite a lot

apps/publish.py
mike-sul: corrected

apps/target_manager.py
mike-sul: fixed

mike-sul:
> looks good. I just found two grammar nits. I usually don't like code that requires such extensive commenting, but in this case, i think it helps quite a lot

ok, I think I/we will be improving the given functionality in the future anyway, so will be able to find out a compact yet informative way to comment it. Also, I think it would be great if we bring the given functionality and this one https://github.com/foundriesio/ci-scripts/blob/master/lmp/customize-target.py#L49 to a common denominator.

mike-sul:
@doanac Is it good to go now?

doanac:
yep. go for it.

PR #97, opened 2020-11-05
lmp: clean up public jobs

ricardosalveti:
- drop ota / ptest build for rpi3-64
- drop supported / other from build names (all valid targets are
supported)
- sync machine list with the supported list from lmp

### DIFF
--- a/lmp/jobserv.yml
+++ b/lmp/jobserv.yml
@@ -11,27 +11,8 @@ triggers:
         refs/heads/master
       OTA_LITE_TAG: postmerge
     runs:
-      # supported images that have OTA
-      - name: supported-{loop}
-        container: hub.foundries.io/lmp-sdk
-        host-tag: amd64-osf
-        loop-on:
-          - param: MACHINE
-            values:
-              - raspberrypi3-64
-        triggers:
-          - name: ota-{loop}
-          - name: ptest-build-{loop}
-        params:
-          IMAGE: lmp-base-console-image
-        script-repo:
-          name: fio
-          path: lmp/build.sh
-        persistent-volumes:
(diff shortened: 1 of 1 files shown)


doanac:
@ricardosalveti - this syntax looks correct, but the schema validator is failing. i think the bug's on my end. i'll take a look

doanac:
I get sick of people complaining about YAML. I kind of like it. But here's why they hate it:
~~~
 - name: {loop}
~~~

Causes a schema validation error like:
~~~
- Value '{'loop': None}' is not of type 'str'. Path: '/triggers/0/runs/1/name'
~~~

The yaml parser is taking `{loop}` not as the string you need, but as a dict `{loop: None}`. So what you need in these new runs are entries like:
~~~
 - name: "{loop}"
~~~

the surrounding quotes fix it.

ricardosalveti:
> I get sick of people complaining about YAML. I kind of like it. But here's why they hate it:
> 
> ```
>  - name: {loop}
> ```
> 
> Causes a schema validation error like:
> 
> ```
> - Value '{'loop': None}' is not of type 'str'. Path: '/triggers/0/runs/1/name'
> ```
> 
> The yaml parser is taking `{loop}` not as the string you need, but as a dict `{loop: None}`. So what you need in these new runs are entries like:
> 
> ```
>  - name: "{loop}"
> ```
> 
> the surrounding quotes fix it.

Lol, nice, so -{loop} would work (assuming yaml identifies the type by the initial char), but {loop} wouldn't.

Nice catch.

ricardosalveti:
Pushed an updated rev with the suggested changes.

doanac (approved):

PR #98, opened 2020-11-11
assemble: fix loop device id

ricardosalveti:
Use correct matching loop device id while parsing the output of losetup
-a, otherwise it will just try to use the first loop device available to
the system.

### DIFF
--- a/assemble.py
+++ b/assemble.py
@@ -37,8 +37,8 @@ class WicImage:
         out = cmd('losetup', '-a', capture=True).decode()
         for line in out.splitlines():
             if self._path in line:
-                self._loop_device = out.split(':', 1)[0]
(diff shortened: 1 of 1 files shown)


assemble.py
```diff
@@ -37,8 +37,8 @@ def __enter__(self):
         out = cmd('losetup', '-a', capture=True).decode()
         for line in out.splitlines():
             if self._path in line:
-                self._loop_device = out.split(':', 1)[0]
-                self._wic_device = out.split(':', 1)[0] + 'p2'
+                self._loop_device = line.split(':', 1)[0]
```
mike-sul, line 40: good catch.

mike-sul (approved):

PR #101, opened 2020-11-11
lmp: Correctly mark currently installed Target

mike-sul:
Mark correctly the currently installed Target. `is_current` attribute should be an attribute of Target not a collection of Targets.


doanac (approved):

ricardosalveti:
Nice, thanks for the fix, let me test this locally now.

ricardosalveti:
Yup, this fixes it, thanks!

PR #106, opened 2020-12-03
containers: Pull in updates for compose-publish

doanac:
This allows us to take service lines like:

 foo:
   image: hub.foundries.io/${FACTORY-andy}/foo

and evaluate them properly. It also pulls in an update from
compose-publish that allows us to get the sha from a file rather than
trying to scan stdout for it.

### DIFF
--- a/apps/apps_publisher.py
+++ b/apps/apps_publisher.py
@@ -2,10 +2,14 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import logging
-from helpers import cmd as cmd_exe
+import os
+from tempfile import NamedTemporaryFile
+
+from expandvars import expandvars
 
 from apps.compose_apps import ComposeApps
 from apps.docker_registry_client import DockerRegistryClient
+from helpers import cmd as cmd_exe
 
 
 logger = logging.getLogger(__name__)
@@ -40,7 +44,8 @@ class AppsPublisher:
     def __tag(self, app: ComposeApps.App, tag: str):
         changed = False
         for _, service_cfg in app.services():
-            image_url = service_cfg['image']
+            image_url = expandvars(service_cfg['image'])
+            logger.info('Service image url: %s', image_url)
             if not image_url.startswith(self._image_base_url):
                 # TODO: verify if URL points to some other factory
                 continue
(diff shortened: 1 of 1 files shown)


doanac:
this requires - https://github.com/foundriesio/lmp-images-tools/pull/7

apps/apps_publisher.py
```diff
@@ -67,13 +71,6 @@ def __publish(self, app: ComposeApps.App, tag: str):
...
-        # The publish command produces output like:
-        # = Publishing app...
-        # |-> app:  sha256:fc73321368c7a805b0f697a0a845470bc022b6bdd45a8b34
-        # |-> manifest:  sha256:e15e3824fc21ce13815aecb0490d60b3a32
-        # We need the manifest sha so we can pin it properly in targets.json
-        needle = b'|-> manifest:  sha256:'
-        sha = out[out.find(needle) + len(needle):].strip()
-        app_hashed_uri = app_base_url + '@sha256:' + sha.decode()
-        return app_hashed_uri
+        cmd_exe(self._publish_tool, '-d', '/tmp/sha', self._app_tagged_url, cwd=app.dir)
```
mike-sul, line 74: I would use something from `tempfile` to avoid usage of hardcoded '/tmp', there is no guarantee that a user running this script has permission to write into this directory (I had this issue when was running tests under tests locally).

apps/apps_publisher.py
```diff
@@ -67,13 +71,6 @@ def __publish(self, app: ComposeApps.App, tag: str):
...
+        with open('/tmp/sha') as f:
+            return app_base_url + '@' + f.read().strip()
```
mike-sul, line 76: As far as I understand this https://github.com/foundriesio/compose-publish/commit/ff9495db64e42a5523f020f21143755721b28c19 dumps a digest in the format we expect here, i.e. `sha256:<hash>`?

mike-sul (commented):
Frankly speaking, I really don't understand the value of using a separate executable for the pinning functionality. It would be much simpler just to implement it in Python in the scope of ci-scripts, it would simplify testing and code support/maintanence.
It's not related to the given PR at all, and not critical atm, I guess this is something that we can put in our backlog.

doanac:
I've force pushed fixes that have this working with a tempfile properly. i'm running it in my factory and all is well.

mike-sul (approved):

PR #107, opened 2020-12-07
lmp/build: do not fail build if extra artifact does not exist

EmbeddedAndroid:
Ran into this issue when building two platforms that each needed to archive an extra artifact. The EXTRA_ARTIFACTS var can be found below:

EXTRA_ARTIFACTS: "flashlayouts-stm32mp1-disco.tar.gz flashlayouts-stm32mp1-eval.tar.gz"

### DIFF
--- a/apps/apps_publisher.py
+++ b/apps/apps_publisher.py
@@ -51,7 +51,8 @@ class AppsPublisher:
                                 .format(app.name, service_name))
             image_url = expandvars(image_url_template)
             logger.info('Service image url: %s', image_url)
-            if not image_url.startswith(self._registry_host) or image_url.startswith(self._registry_host + "/lmp/"):
+            if not image_url.startswith(self._registry_host) or \
+                    (self._factory != 'lmp' and image_url.startswith(self._registry_host + "/lmp/")):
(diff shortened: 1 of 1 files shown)


lmp/build.sh
```diff
@@ -133,7 +133,7 @@ if [ -d "${archive}" ] ; then
 	# Handle user provided extra artifacts
 	if [ ! -z "${EXTRA_ARTIFACTS}" ]; then
 		for extra_file in ${EXTRA_ARTIFACTS}; do
-			cp ${DEPLOY_DIR_IMAGE}/${extra_file} ${archive}/
+			cp ${DEPLOY_DIR_IMAGE}/${extra_file} ${archive}/ || true
```
doanac, line 136: Maybe we should do something like:
~~~
if [ -f ${DEPLOY_DIR_IMAGE}/${extra_file} ] ; then
    cp ${DEPLOY_DIR_IMAGE}/${extra_file} ${archive}/
else
  echo "Skipping archiving ${extra_file} - file does not exist"
fi
~~~

EmbeddedAndroid: Yeah, probably best to log a message since the user is explicitly asking for this file. I'll update the PR.

PR #108, opened 2020-12-08
lmp/bb-config: always disable sstate mirrors

ricardosalveti:
Minor one, but noticed when comparing master x devel (only devel had SSTATE_MIRROR disabled).

### DIFF
--- a/lmp/bb-config.sh
+++ b/lmp/bb-config.sh
@@ -143,12 +143,10 @@ INHERIT += "archiver"
 COPYLEFT_RECIPE_TYPES = "target"
 ARCHIVER_MODE[src] = "original"
 ARCHIVER_MODE[diff] = "1"
-EOFEOF
 
-if [ $(ls ../sstate-cache | wc -l) -ne 0 ] ; then
-	status "Found existing sstate cache, using local copy"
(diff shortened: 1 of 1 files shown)


lmp/bb-config.sh
```diff
@@ -143,12 +143,10 @@ INHERIT += "archiver"
...
 ARCHIVER_MODE[diff] = "1"
-EOFEOF
 
-if [ $(ls ../sstate-cache | wc -l) -ne 0 ] ; then
-	status "Found existing sstate cache, using local copy"
-	echo 'SSTATE_MIRRORS = ""' >> conf/auto.conf
-fi
+# no sstate mirrors, using local copy
+SSTATE_MIRRORS = ""
+EOFEOF
```
doanac, line 150: while we are looking at this, should we also consider trying this:

 https://github.com/doanac/ci-scripts/commit/9504b573260b91a4e5506328a255bc1eeeff4150 ?

ricardosalveti: Sure, that would be better as I believe you already tested that path.

Mine was just to remove something that wasn't really being used.

Will close this one then, and let you push the other change.

PR #109, opened 2020-12-08
lmp/bb-config: drop exports from local.conf

ricardosalveti:
Drop the need for exporting variables from local.conf as that causes a
massive bad side effect on caching.

Move variables to be provided as parameters for the customize-target
script instead, and drop variables that are not used by LMP.

### DIFF
--- a/lmp/bb-config.sh
+++ b/lmp/bb-config.sh
@@ -2,7 +2,7 @@
 
 HERE=$(dirname $(readlink -f $0))
 source $HERE/../helpers.sh
-require_params MACHINE IMAGE
+require_params MACHINE IMAGE OTA_LITE_TAG
 
 OSTREE_BRANCHNAME="${OSTREE_BRANCHNAME-lmp-localdev}"
 SOTA_CLIENT="${SOTA_CLIENT-aktualizr}"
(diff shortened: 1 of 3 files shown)


doanac (approved):
nice work. thanks!

lmp/bb-config.sh
```diff
@@ -47,7 +47,7 @@ OSTREE_BRANCHNAME = "${MACHINE}-${OSTREE_BRANCHNAME}"
 GARAGE_SIGN_REPO = "/tmp/garage_sign_repo"
 GARAGE_TARGET_VERSION = "${H_BUILD}"
 GARAGE_TARGET_URL = "https://ci.foundries.io/projects/${H_PROJECT}/builds/${H_BUILD}"
-GARAGE_CUSTOMIZE_TARGET = "${HERE}/customize-target.sh ${GARAGE_CUSTOMIZE_TARGET_PARAMS}"
+GARAGE_CUSTOMIZE_TARGET = "${HERE}/customize-target.sh ${FACTORY} ${OTA_LITE_TAG} ${GARAGE_CUSTOMIZE_TARGET_PARAMS}"
```
mike-sul, line 50: It could be also added at line #21.

mike-sul (approved):
That's great not just in terms of improving sstate cache hit, but also in terms of code quality, it's much easier to understand what's going if parameters are passed explicitly. I've been doing similar in other parts of ci-scripts that are related to the preloading.

lmp/bb-config.sh
ricardosalveti: I saw that but we actually can't because that line is given in a way that it will keep the ${} format so it can later be used by bitbake, and these 2 variables are not exported to bitbake.

Which is why I added here, is it could be translated from shell to local.conf directly on execution.

mike-sul: :+1:

mike-sul: FACTORY -> LMP_DEVICE_FACTORY
for the tag we can set OTA_LITE_TAG = "${OTA_LITE_TAG}" in local.conf.

`export` is really required for those variables that are needed by python scripts called from recipes.

Anyway, this is really not important, it's even better, iirc, I did this trick in the first place because some of the variables are not defined at this stage (${IMAGE_BASENAME} ${TARGET_ARCH} at least back then).

doanac:
@ricardosalveti - builds in my factory. 2nd build it 98%. nice work

PR #110, opened 2020-12-09
Initial prototype of OSTree-based Compose Apps

mike-sul:


mike-sul:
@doanac effectively this is the implementation of item #1 of the implementation plan https://docs.google.com/document/d/1lQs5Sas0WwVFzcaH4v-UYelgvaYn6_PBkEGs-Ndop5I/edit#heading=h.e4bsxuh2k3zs

I think it can even be deployed without impacting existing devices and customers, just the container job (publish run) will take a bit longer and an ostree repo will be used for the preloading.

apps/ostree_store.py
```diff
@@ -0,0 +1,76 @@
...
+        tag = '_'.join(target.tags)
+        branch = self.branch(tag, target.platform)
+        hash = self._repo.commit(images_dir, branch)
+        self._push_to_treehub(hash)
+
+        return hash
+
+    def _remove_and_index_character_devices(self, tree_to_commit):
+        char_dev_list = subprocess.check_output(['find', tree_to_commit, '-type', 'c'])
+        print(char_dev_list)
```
doanac: maybe use a logger for this?

apps/ostree_store.py
```diff
@@ -0,0 +1,76 @@
...
+        for dev_file in char_dev_list.splitlines():
+            os.remove(dev_file)
+
+    def _push_to_treehub(self, hash):
+        push = subprocess.Popen(['garage-push', '--repo', self._repo.dir,
+                                 '--credentials', self._cred_arch, '--ref', hash])
+
+        push.wait()
+        push.terminate()
```
doanac: is there something subtle here to explain why not subprocess.check_call?

apps/publish.sh
```diff
@@ -95,7 +97,18 @@ status "Publishing apps; version: ${APPS_VERSION}, Target tag: ${TARGET_TAG}"
...
+
+
+/usr/local/bin/dind "${HERE}/publish_apps_to_ostree.py" \
+  --factory "${FACTORY}" \
+  --token "$(cat "${SECRETS}/osftok")" \
+  --cred-arch "${CREDS_ARCH_UPDATED}" \
+  --targets-file "${TUF_REPO}/roles/unsigned/targets.json" \
+  --targets-to-publish "${ARCHIVE}/targets-created.json" \
+  --fetch-dir "${PRELOAD_DIR}" \
+  --repo-dir "${OSTREE_REPO_DIR}"
```
doanac: Should we make this configurable/optional for now? Or maybe we go all-in and avoid making us all do factory-config changes to try?

apps/publish_apps_to_ostree.py
```diff
@@ -0,0 +1,59 @@
...
+        with open(args.targets_file, 'r') as f:
+            all_targets = json.load(f)
+            for target_name, target_json in targets.items():
+                all_targets['targets'][target_name] = target_json
+
+        with open(args.targets_file, 'w') as f:
+            json.dump(all_targets, f, indent=2)
+
+    except Exception as exc:
+        logging.error('Failed to fetch Target apps and images: {}'.format(exc))
```
doanac: maybe logging.exception? i'm finding logging.error is hard to debug for lots of exceptions like "KeyError".

assemble.py
```diff
@@ -117,23 +118,16 @@ def copy_container_images_to_wic(target: FactoryClient.Target, factory: str, app
...
-            logger.info('Checking out Apps from an ostree repo; src={}, dst={}'.
-                        format(wic_image.compose_apps_tree, wic_image.compose_apps_root))
-            dst_apps_tree_repo.checkout(target.apps_sha, '/apps', wic_image.compose_apps_root)
-            logger.info('Checking out Apps\' container images from an ostree repo; src={}, dst={}'.
-                        format(wic_image.compose_apps_tree, wic_image.docker_data_root))
-            dst_apps_tree_repo.checkout(target.apps_sha, '/images', wic_image.docker_data_root)
+            target_app_store.copy_and_checkout(target, wic_image.compose_apps_tree,
+                                               wic_image.compose_apps_root, wic_image.docker_data_root)
 
             wic_image.update_target(target)
```
doanac, line 132: looks like a p.tick got removed. progress should self-correct I think, but it might be a little jittery

apps/publish.sh
```diff
@@ -36,7 +36,8 @@ MACHINES=${MACHINES-""}
 PLATFORMS=${MANIFEST_PLATFORMS_DEFAULT-""}
 
 APP_IMAGES_ROOT_DIR="${APP_IMAGES_ROOT_DIR-/var/cache/bitbake/app-images}"
-OSTREE_REPO_DIR="${APP_IMAGES_ROOT_DIR}/ostree-repo"
+#OSTREE_REPO_DIR="${APP_IMAGES_ROOT_DIR}/ostree-repo"
+OSTREE_REPO_DIR="${OSTREE_REPO_DIR-$(mktemp -d -p ${HOME})}"
```
doanac, line 40: can just delete this commented out line. I assume using a local volume speeds things up because it skips overlayfs or something?

doanac (commented):
looking good.

apps/ostree_store.py
mike-sul: that's debugging leftover, I will just remove it.

apps/ostree_store.py
mike-sul: @doanac It looks like you've reviewed some outdated/intermediate version, I had to do forced push what might confuse you.

apps/publish.sh
mike-sul: It doesn't require any changes to `factory-config` to try it out, those params have default definition at the beginning of the script.

apps/publish_apps_to_ostree.py
mike-sul: ok

assemble.py
mike-sul: there are 'ticks' in copy_container_images_to_wic, or do you mean specifically at this location, just after copy_and_checkout?

apps/publish.sh
mike-sul: ${APP_IMAGES_ROOT_DIR}/ostree-repo refers to NFS and committing to an ostree repo located on NFS is veryyy slow.
Sure, I will remove it.
OSTREE_REPO_DIR="${OSTREE_REPO_DIR-$(mktemp -d -p ${HOME})}" - this actually points to container write layer managed by overlayfs, it doesn't cause much trouble, but ideally, it would be great if jobserve worker adds `--tmpfs` so the code running in a container could store something directly on a host filesystem. In this case, I could do "checkout with hardlinks", fetch and then commit back with `--link-checkout-speedup` which helps to speed up the whole process.

mike-sul:
I need to rebase it on the latest master and then retest it again.

apps/publish.sh
doanac: but since its not optional, this means that *all* factories will start producing this new content once merged, correct?

mike-sul: yes, what should not impact their existing workflow. We can make it optional, but then another question arises - how to do preload?

assemble.py
mike-sul: done

apps/publish.sh
mike-sul: so, I am making it OFF by default, just do exactly as it does now, COMPOSE_APP_USE_OSTREE set in factory-config.yml turns the ostree usage ON.
The only question is in regard to 'fioctl target image' command since it doesn't have access to factory-config.yml.

PR #112, opened 2020-12-14
Preload apps enhancements

doanac:
This allows APP_SHORTLIST to work in LMP builds and adds the ability to preload compose app content.


doanac:
@mike-sul - I'm about 90% sure you aren't going to like the way i'm storing compose apps. I just went with the least invasive way I could find to build the change though.

One interesting thing I hit was how lmp/preload-app-images uses positional arguments. I had to use named arguments so that ci-scripts won't break meta-lmp. We can/should still update meta-lmp to call this explicitly, but it gives us some wiggle room.

apps/publish.sh
```diff
@@ -41,6 +41,8 @@ PLATFORMS=${MANIFEST_PLATFORMS_DEFAULT-""}
 require_params FACTORY ARCHIVE TARGET_TAG
 #-- END: Input params
 
+apk add tar
```
doanac, line 44: the way i call tar requires the gnu version of tar and not busybox-tar. We can update lmp-image-tools if this PR is okay. but i have a feeling you'll want to completely overhaul this approach, so i left it like this to keep it testable in my factory.

mike-sul:
> One interesting thing I hit was how lmp/preload-app-images uses positional arguments. I had to use named arguments so that ci-scripts won't break meta-lmp. We can/should still update meta-lmp to call this explicitly, but it gives us some wiggle room.

That's OK, I do the same for the ostree based compose apps. Once changes (optional argument) to ci-scripts are merged to master I would say that it would be better to make an explicit call in meta-lmp.

Why do we need this shortlisting at all. I mean if a user doesn't need a certain app why not just remove it from containers.git.

mike-sul:
> I just went with the least invasive way I could find to build the change though.

I don't mind taking into account that sooner or later it will be replaced with the ostree-based compose apps (preloading of both apps and images is implemented there). 
Maybe it makes sense to put the "target" directory into tar and then extract `apps` and `images` directory by separate tar calls. 

`work-dir/app-images/fetch-dir/intel-corei7-64-lmp-235/
├── apps
│   ├── app-05
│   └── app-07
└── images
    ├── image
    └── overlay2
`

archive: tar -cf preload.tar -C work-dir/app-images/fetch-dir/intel-corei7-64-lmp-235/ .  (TargetAppsFetcher::target_dir)

extract images to the docker datat root: tar -xvf preload.tar --strip-components=2 -C tmp/docker/ ./images
extract apps to the compose-apps dir: tar -xvf preload.tar --strip-components=2 -C tmp/compose-apps/ ./apps

But, yes, your approach is less invasive, the approach I outline require falling back to the previous approach if the archive is in th e old/current format and contains just `images`.
Another simple and less invasive approach would be just creating another tar archive for `apps` so during preloading if it exists then it's copied to a system image if not just preload images.

doanac:
> Why do we need this shortlisting at all. I mean if a user doesn't need a certain app why not just remove it from containers.git.

I have a setup where I'm building 3 apps. By default I want 2 of those enabled, but we need the option to enable the 3rd instead.

doanac:
> Another simple and less invasive approach would be just creating another tar archive for apps so during preloading if it exists then it's copied to a system image if not just preload images.

I thought that would make a lot of sense. I was having a hard time understanding what all I needed to add/change to make that work. any pointers?

doanac:
> Anyway, I am not sure that apps shortlisting will work for the preloading in the scope of LmP build, since the lmp preloader just simply copies what was fetched/dumped during the corresponding container build. So, if shortlisting was set ON during a container build then corresponding shortlisted apps will be dumped and the archive file will be named as for example 155f56cd44de89fedf7c819985549754b4a2383b-arm64-app-05.tar. It means that during the LmP preloading the source archive filename should be composed correspondingly - take into account args.app_shortlist but I don't see any usage of it in 191658a.

ah - i should remove the apps_shortlist stuff from preload-app-images. i guess i got ahead of myself on that.

mike-sul (commented):
I suppose this becomes redundant because of this PR https://github.com/foundriesio/ci-scripts/pull/115

doanac:
i'd forgotten that i'd done this.

PR #113, opened 2020-12-16
lmp: add custom debug config log and publish debug/warning

ricardosalveti:
Add logic for handling a custom config log that improves the standard stdout output by adding timestamp and redirects the warn/debug information into 2 files, which are later uploaded as part of the build process.

You can see how it looks like by checking https://ci.foundries.io/projects/rsalveti-v77/lmp/builds/9/n1sdp/artifacts/console.log.

### DIFF
--- a/lmp/bb-build.sh
+++ b/lmp/bb-build.sh
@@ -6,6 +6,6 @@ require_params IMAGE
 
 source setup-environment build
 
-bitbake ${IMAGE}
-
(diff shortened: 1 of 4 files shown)


ricardosalveti:
Will merge so we can do some more debugging on this, tested on rsalveti-v77.

doanac (commented):
really cool. thanks

PR #114, opened 2020-12-16
preload: preload compose apps along with their images

mike-sul:
Preload Compose Apps's content in addition to preloading of their container images.


mike-sul:
@doanac It's not exactly what we discussed but it looks simpler compared to having two separate archive files for images and apps.

apps/target_apps_store.py
```diff
@@ -70,26 +70,55 @@ def images_size(self, target: FactoryClient.Target):
 
         return app_image_size
 
-    def copy(self, target: FactoryClient.Target, dst_dir: str):
+    def copy(self, target: FactoryClient.Target, dst_images_dir: str, dst_apps_dir: str):
+        if not self.exist(target):
```
doanac, line 74: this is for backwards compatibility, correct? This conditional branch can eventually be removed?

apps/target_apps_store.py
```diff
@@ -70,26 +70,55 @@ def images_size(self, target: FactoryClient.Target):
...
+        if os.path.exists(dst_apps_dir):
+            # wic image was populated by container images data during LmP build
+            # let's remove it and populate with the given images data
+            logger.info('Removing existing preloaded compose apps from the system image')
+            shutil.rmtree(dst_apps_dir)
+
+        os.makedirs(dst_apps_dir)
+        logger.info('Copying Target Apps; Target: {}, Source: {}, Destination: {}'
+                    .format(target.name, app_image_tar, dst_apps_dir))
+        subprocess.check_call(['tar', '--strip-components=2', '--exclude=./images',
```
doanac, line 109: stinks there's not a "--include" option. that would make it a little more future proof

doanac (commented):
looks good.

apps/target_apps_store.py
doanac: do we need to update lmp-image-tools also?

mike-sul: > do we need to update lmp-image-tools also?

no, it works as it is, I verified key use-cases in my factory.

mike-sul: > stinks there's not a "--include" option. that would make it a little more future proof

I think there is another way to achieve "--include effect", will try it tomorrow.

apps/target_apps_store.py
mike-sul: yes

apps/target_apps_store.py
mike-sul: updated to make it "include" not exclude.

doanac (approved):
thanks!

PR #115, opened 2021-01-12
lmp: Improve app preloading logic

doanac:
There were a few problems with app short list functionality:

bb-config.sh
 * we weren't passing the list to the preload script

preload-app-images
 * It had no notion of where /var/sota/compose-apps was. This change
   fixes things without having to update meta-lmp.

I also changed the logging of exceptions to include the stack trace
which helps debug unexpected errors greatly.

Lastly, I added a "fail-on-error" option that allows OE builds to fail
if you want. This helps debug quite a bit.


lmp/preload-app-images
```diff
@@ -68,7 +72,9 @@ if __name__ == '__main__':
     try:
         preload_app_images(args.installed_versions, args.app_shortlist, args.images_root_dir, args.apps_root_dir, args.dst_dir)
     except Exception as exc:
-        logger.warning('Failed to preload Apps images: {}'.format(exc))
+        logger.exception('Failed to preload Apps images: {}'.format(exc))
+        if args.fail_on_error:
```
mike-sul, line 76: IIUC the `fail_on_error` is False by default, iirc that's what Mike Scott said is preferable from a user perpsective, OE build should succeed regardless of the preload success.

mike-sul (approved):

PR #116, opened 2021-02-02
A forward-compatible change for a new format of GET /ota/factories/<factory>/targets/

vkhoroz:


doanac (commented):
seems fine, but I think this is up to @mike-sul

mike-sul (approved):

MiloCasagrande (approved):

PR #117, opened 2021-02-16
Improve error logging

mike-sul:
Yield a meaningful error message if
1. `image` element is not found in a compose app service configuration (docker-compose.yml:service);
2. image URI refers to an image that does not belong to the given Factory.

### DIFF
--- a/apps/apps_publisher.py
+++ b/apps/apps_publisher.py
@@ -43,8 +43,12 @@ class AppsPublisher:
 
     def __tag(self, app: ComposeApps.App, tag: str):
         changed = False
-        for _, service_cfg in app.services():
-            image_url = expandvars(service_cfg['image'])
+        for service_name, service_cfg in app.services():
+            image_url_template = service_cfg.get('image')
+            if not image_url_template:
+                raise Exception('Mandatory element `image:` is missing in the app service config; app: {}, service: {}'
+                                .format(app.name, service_name))
+            image_url = expandvars(image_url_template)
             logger.info('Service image url: %s', image_url)
             if not image_url.startswith(self._image_base_url):
                 # TODO: verify if URL points to some other factory
--- a/apps/apps_publisher.py
+++ b/apps/apps_publisher.py
@@ -17,8 +17,9 @@ logger = logging.getLogger(__name__)
 
 class AppsPublisher:
     def __init__(self, factory, publish_tool: str, registry_host=DockerRegistryClient.DefaultRegistryHost):
-        self._publish_tool = publish_tool
         self._factory = factory
+        self._publish_tool = publish_tool
+        self._registry_host = registry_host
 
         self._image_base_url = '{}/{}'.format(registry_host, self._factory)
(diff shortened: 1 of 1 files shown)


apps/apps_publisher.py
```diff
@@ -50,9 +51,12 @@ def __tag(self, app: ComposeApps.App, tag: str):
                                 .format(app.name, service_name))
             image_url = expandvars(image_url_template)
             logger.info('Service image url: %s', image_url)
-            if not image_url.startswith(self._image_base_url):
-                # TODO: verify if URL points to some other factory
+            if not image_url.startswith(self._registry_host):
```
mwasilew, line 54: How do you refer to dockerhub images after this change?

mike-sul: As usually, the change doesn't impact it. The given business logic tags only URLs of images that are hosted in Foundries Docker registry, if an image URL refers to some 3rd party registry/hub then we don't need to do anything, so the code just call 'continue' in this case.

mwasilew: ok, so if I declare it as:
```
image: foo:latest
```
it will default to dockerhub and it will still work?

mike-sul: yes, provided that 'foo:latest' is hosted in the docker hub.

apps/apps_publisher.py
```diff
@@ -43,12 +44,19 @@ def tag(self, apps: ComposeApps, version: str):
...
+            image_url = expandvars(image_url_template)
             logger.info('Service image url: %s', image_url)
-            if not image_url.startswith(self._image_base_url):
-                # TODO: verify if URL points to some other factory
+            if not image_url.startswith(self._registry_host):
+                # image url points to non-Foundries docker registry
                 continue
+            if not image_url.startswith(self._image_base_url):
+                raise Exception('Image url refers to an image that does not belong to the given Factory;'
+                                ' url: {}, factory: {}'.format(image_url, self._factory))
```
doanac, line 59: I think this means you could do something like:
~~~
 image: hub.foundries.io/lmp/fiotest:postmerge
~~~

mike-sul: or if a build is for msul-dev01 factory and I am referring to an image of msul-dev02 (we can tag only those images that we build).

doanac: we need to allow customers to use the "lmp" factory containers.

mike-sul: allow to use but not tag the image? so just skip/continue if the logic finds URL referring to the "lmp" factory container?

mike-sul:
@doanac I forced pushed the change to allow using containers from the public (lmp) Factory

`2021-02-17 08:11:08,408 INFO: apps_publisher: Service image url: hub.foundries.io/lmp/fiotest:postmerge
2021-02-17 08:11:08,411 INFO: apps_publisher: Publishing Apps..`
https://ci.foundries.io/projects/msul-dev01/lmp/builds/734/publish-compose-apps/console

doanac (approved):

PR #118, opened 2021-02-24
helpers: Wrap repo-sync with timeout and retry

doanac:
We are seeing occasional hangs where repo-sync calls git which calls
some git-fetch command that seems to hang forever. The root issue
is bad international networking. This change attempts to deal with it
by:

 * placing a reasonable timeout on a repo-sync command
 * retrying a few times if the sync failed due to a timeout

### DIFF
--- a/helpers.sh
+++ b/helpers.sh
@@ -32,7 +32,18 @@ function repo_sync {
 		git config --global http.https://${domain}.extraheader "$(cat /secrets/git.http.extraheader)"
 	fi
 	run repo init --no-clone-bundle -u $* ${REPO_INIT_OVERRIDES}
-	run repo sync
+	for i in $(seq 4); do
+		run timeout 4m repo sync && break
+		if [ $? -eq 124 ] ; then
+			msg="Command timed out"
+			if [ $i -ne 4 ] ; then
+				msg="${msg}, trying again"
+			fi
+			status ${msg}
+		else
+			exit $?
+		fi
+	done
 	if [ -d "$archive" ] ; then
 		status "Generating pinned manifest"
 		repo manifest -r -o $archive/manifest.pinned.xml


doanac:
@ricardosalveti @mike-sul @kostyapenkov  - I'm not particularly proud of this change. However, in the absence of guaranteed international networking, I think repo-sync logic needs some retry/timeout protections.

5 minutes almost seems too high - JobServ is going to print the "cmd seems hung" message around that point. So I wonder if I should make it 4 just to avoid the noise of that? 

Testing in my factory here: https://ci.foundries.io/projects/andy-corp/lmp/builds/247/

Its not exactly trivial to force the failure we've seen. So I'm not 100% sure this fixes.

kostyapenkov:
@doanac I'm also not sure that retires will be helpful, but absolutely agree that builds should fail after some timeout and do not wait repo sync for ages.
Unfortunately, it is noting for tuning in repo sync: https://source.android.com/setup/develop/repo#sync (just increase number of threads to sync faster, but it doesn't related to hung issue).

Some historical data: average `repo sync` time for our builders is 70 seconds, and maximum is 130 seconds when networking works "as usual". Average is about 1400s and maximum 4900s when networking is bad. Value for timeout 3-4 minutes fits into average and shouldn't introduce any additional issues.

helpers.sh
```diff
@@ -32,7 +32,18 @@ function repo_sync {
...
 	run repo init --no-clone-bundle -u $* ${REPO_INIT_OVERRIDES}
-	run repo sync
+	for i in $(seq 4); do
+		run timeout 5m repo sync && break
+		if [ $? -eq 124 ] ; then
+			msg="Command timed out"
+			if [ $i -ne 4 ] ; then
+				msg="${msg}, trying again"
+			fi
+			status Command timed out, trying again
```
kostyapenkov, line 42: I guess it should be `status ${msg}`

kostyapenkov:
> I'm also not sure that retires will be helpful

But might be...
until `repo sync` is running inside the same instance (container), each retry will try to fetch only new data, and establish new TCP connection.

helpers.sh
ricardosalveti: Yeah, not sure why we need this additional if if 124 means a timeout, you could just have the status line as it will be aborted correctly coming from the previous for.

ricardosalveti:
> > I'm also not sure that retires will be helpful
> 
> But might be...
> until `repo sync` is running inside the same instance (container), each retry will try to fetch only new data, and establish new TCP connection.

Yeah, a new connection might have it fixed.

+1 for 4 minutes timeout.

doanac:
force pushed with 4 minute timeout and status message fix.

ricardosalveti (approved):
LGTM

PR #122, opened 2021-03-23
apps: Handle publishing for docker-compose.yaml files

doanac:
We currently only look for `docker-compose.yml` files. This can be
confusing to a user that creates a `docker-compose.yaml`.

This isn't the cleanest approach, but our code expects
docker-compose.yml in a couple of places. This fixes things in the
least invasive way.

### DIFF
--- a/apps/compose_apps.py
+++ b/apps/compose_apps.py
@@ -21,7 +21,11 @@ class ComposeApps:
 
         @staticmethod
         def is_compose_app_dir(app_dir):
-            return os.path.exists(os.path.join(app_dir, ComposeApps.App.ComposeFile))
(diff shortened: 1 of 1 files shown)


doanac:
@mike-sul - this should fix the problem @ricardosalveti hit today. its not the cleanest approach, so feel free to reject and do it a different way. I was just aiming for a tiny diff.

mike-sul (approved):
Maybe it would be better just to return an error and stop a CI build with an error message so we enforce customers to keep their compose files nice and tidy.

doanac:
@mike-sul - that's a better idea. I've force pushed that idea.

mike-sul (approved):

mike-sul:
@doanac I guess that's fine to merge it.

PR #131, opened 2021-05-31
Add ability to bundle encrypted apps into the system image

doanac:
This is a 3 part change:

* Passing short-list as a parameter to target-store makes no sense. The target store names the artifact using the the Target.shortlist attribute.
* Target fetching could be simplified and made more consistent
* Support of encrypted apps.

In the end you can run a command like `fioctl secrets update app-encrypted-key=SuperSecretValue` and then have a factory config ref option like
~~~
refs/heads/devel:
      params:
        APP_SHORTLIST: "shellhttpd"
        APP_ENCRYPTED: "fiotest"
~~~

This will produce a system image with shellhttpd preloaded and included in the initial Target aklite runs. There will also be a self extracting file `/var/sota/extract-encrypted-apps SuperSecretValue`. At this point the containers and compose app will be in place.

### DIFF
--- a/apps/target_apps_store.py
+++ b/apps/target_apps_store.py
@@ -98,7 +98,7 @@ class ArchiveTargetAppsStore(TargetAppsStore):
 
         return app_image_size
 
-    def _copy_encrypted(self, app_img: str, dst_images_dir: str, encrypt_apps: str):
+    def _copy_encrypted(self, app_img: str, dst_images_dir: str, encryption_key: str):
         path = os.path.join(dst_images_dir, 'extract-encrypted-apps')
         logger.info('Creating self extracting encrypted file: %s', path)
         with open(path, 'wb') as f:
@@ -106,7 +106,7 @@ class ArchiveTargetAppsStore(TargetAppsStore):
             f.write(EXTRACTION_SCRIPT.encode())
             f.flush()
             subprocess.check_call(['openssl', 'enc', '-pbkdf2', '-e', '-aes256',
-                                   '-in', app_img, '-k', encrypt_apps], stdout=f)
+                                   '-in', app_img, '-k', encryption_key], stdout=f)
 
     def copy(self, target: FactoryClient.Target, dst_images_dir: str, dst_apps_dir: str):
         if not self.exist(target):
--- a/apps/target_apps_store.py
+++ b/apps/target_apps_store.py
@@ -32,6 +32,34 @@ class TargetAppsStore:
     def apps_location(self, target: FactoryClient.Target):
         pass
 
+EXTRACTION_SCRIPT = '''#!/bin/sh
+set -e
+set -o pipefail
+if [ $# -ne 1 ] ; then
+    echo "Usage: $0 <decryption key>"
+    exit 1
+fi
+
+archive=$(grep --text --line-number '# ARCHIVE:$' $0 | cut -d: -f1)
+cp /var/lib/docker/image/overlay2/repositories.json /tmp/orig.json
+echo "Extracting container images"
+tail -n +$((archive + 1)) $0 | openssl enc -pbkdf2 -d -aes256 -k $1 \\
+    | tar -C /var/lib/docker --strip-components=2 -x ./images
+
+python3 -c 'import json; cur = json.load(open("/tmp/orig.json")); new = json.load(open("/var/lib/docker/image/overlay2/repositories.json")); cur["Repositories"].update(new["Repositories"]); json.dump(cur, open("/var/lib/docker/image/overlay2/repositories.json", "w"))'
+
+echo "Extracting compose apps"
+tail -n +$((archive + 1)) $0 | openssl enc -pbkdf2 -d -aes256 -k $1 \\
+    | tar -C /var/sota/compose-apps --strip-components=2 -x ./apps
+
+rm $0
+
+systemctl restart docker
+
+exit 0
+
+# ARCHIVE:
+'''
 
 class ArchiveTargetAppsStore(TargetAppsStore):
     def __init__(self, root_dir):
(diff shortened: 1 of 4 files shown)


assemble.py
```diff
@@ -213,22 +244,28 @@ def get_args():
         p = Progress(len(targets))
         logger.info('Found {} Targets to assemble image for'.format(found_targets_number))
         for target in targets:
-            logger.info('Assembling image for {}, shortlist: {}'.format(target.name, args.app_shortlist))
+            logger.info('Assembling image for %s, shortlist: %s, encrypted: %s',
+                        target.name, args.app_shortlist, args.app_encrypted)
             subprog = Progress(3, p)
             image_file_path = factory_client.get_target_system_image(target, args.out_image_dir, subprog)
+            target.shortlist = args.app_shortlist
```
mike-sul, line 251: Maybe if a new property of Target is introduced, e.g. `Target.enc_shortlist` then we could avoid many 'if' statements at the upper level, it would be just in the scope of Fetcher implementation? (Just guessing :)).

doanac: that way is possible. It requires adding new logic to the fetcher to fetch both things into different locations. And then you have to add a different set of if's into the target-store so that store content and name it correctly. That was my initial approach and I was hitting a lot of friction with the fetcher and store modules. I don't mind trying it again if you think that's the way to go.

mike-sul: >I don't mind trying it again if you think that's the way to go.

No, no need, I am pretty sure I thought it through and I have already forgotten the implementation details :).

apps/target_apps_store.py
```diff
@@ -32,6 +32,34 @@ def copy(self, target: FactoryClient.Target, dst_images_dir: str, dst_apps_dir:
...
+if [ $# -ne 1 ] ; then
+    echo "Usage: $0 <decryption key>"
+    exit 1
+fi
+
+archive=$(grep --text --line-number '# ARCHIVE:$' $0 | cut -d: -f1)
+cp /var/lib/docker/image/overlay2/repositories.json /tmp/orig.json
+echo "Extracting container images"
+tail -n +$((archive + 1)) $0 | openssl enc -pbkdf2 -d -aes256 -k $1 \\
+    | tar -C /var/lib/docker --strip-components=2 -x ./images
```
mike-sul, line 47: Let's hope that a simple file-level merge of existing container images (dockerd's image and layer stores) will work (I've already forgotten details  on this matter :) )

assemble.py
```diff
@@ -140,8 +140,7 @@ def copy_container_images_from_archive_to_wic(target: FactoryClient.Target, app_
     if not target_app_store.exist(target):
         logger.info('Container images have not been found, trying to obtain them...')
         apps_fetcher = TargetAppsFetcher(token, app_preload_dir)
-        apps_fetcher.fetch_target_apps(target, target.shortlist)
-        apps_fetcher.fetch_apps_images()
+        apps_fetcher.fetch_target(target)
```
mike-sul, line 149: Nice catch, I recall that I always wanted to do it but never got to it.

assemble.py
```diff
@@ -134,15 +133,14 @@ def copy_container_images_to_wic(target: FactoryClient.Target, factory: str, ost
...
+                                              wic_image: str, token: str, progress: Progress):
 
     p = Progress(2, progress)
     target_app_store = ArchiveTargetAppsStore(app_image_dir)
-    target.shortlist = apps_shortlist
     if not target_app_store.exist(target):
         logger.info('Container images have not been found, trying to obtain them...')
         apps_fetcher = TargetAppsFetcher(token, app_preload_dir)
-        apps_fetcher.fetch_target_apps(target, apps_shortlist)
+        apps_fetcher.fetch_target_apps(target, target.shortlist)
```
mike-sul: That's great. I think the change won't impact the LmP's preload case since the script uses the Target object attribute to pass a shortlist, too. https://github.com/foundriesio/ci-scripts/blob/master/lmp/preload-app-images#L69.

mike-sul:
I did an initial review, it looks great, quite minimalistic change, will play/test it later today.

IIUC, we need to support it just for the container build and fioct use-cases, supporting of the LmP build use-case is not needed or will be implemented later https://github.com/foundriesio/ci-scripts/blob/master/lmp/preload-app-images? 

Since the mix of clear text and encrypted apps require merging of the dockerd's image and layer stores we might wanna consider making use of `docker save/load` in order to make sure that it's done correctly. According to this https://docs.google.com/document/d/1lQs5Sas0WwVFzcaH4v-UYelgvaYn6_PBkEGs-Ndop5I/edit#heading=h.i5195gytomaf (see p. 2 Images’ metadata merging issue) manual merging of `repository.json` should be enough, although I mentioned one more item in the doc...

doanac:
> IIUC, we need to support it just for the container build and fioct use-cases, supporting of the LmP build use-case is not needed or will be implemented later https://github.com/foundriesio/ci-scripts/blob/master/lmp/preload-app-images?

Nope - i forgot to do that. I'll work on a patch for that.

apps/target_apps_store.py
```diff
@@ -70,16 +98,28 @@ def images_size(self, target: FactoryClient.Target):
 
         return app_image_size
 
+    def _copy_encrypted(self, app_img: str, dst_images_dir: str, encrypt_apps: str):
```
mike-sul, line 101: `encrypt_apps` a bit confusing param name, it's an encryption key, isn't it?

apps/target_apps_store.py
mike-sul: Will the archive itself stay on the filesystem after extraction?

mike-sul: No, it won't. already verified.

doanac:
> Since the mix of clear text and encrypted apps require merging of the dockerd's image and layer stores we might wanna consider making use of docker save/load in order to make sure that it's done correctly.

docker-save doesn't save enough image metadata. You can save by sha256, but then you get:
~~~
# docker images
REPOSITORY   TAG       IMAGE ID       CREATED       SIZE
<none>       <none>    d3cd072556c2   2 weeks ago   1.24MB
~~~

The image data is there, but when you try and run that container image, it does a lookup on the docker registry to then understand it already has the data.

You can save by a tag, but then docker looks up the sha256 to then find out it already has the data.

Either way - its not a totally "saved" image for our use case.

apps/target_apps_store.py
doanac: https://github.com/foundriesio/ci-scripts/pull/131/commits/15580749455d264e0e1b0431e14cbb2001df69f3

mike-sul (approved):

doanac:
we've decided to hold off on this change for now.

PR #133, opened 2021-06-08
apps: build: drop image label to fix manifest sha

mike-scott:
Testing shows that using --label="..." with a different value
each time you run a build will generate a different manifest sha
when pushed to a registry.

This label is used to identify the build a particular image was
created in.  For now, let's drop the label and investigate a
better way to identify images on hub.foundries.io (if at all).


mike-sul:
I am not sure if the given label is used at all, so I think that's great to drop it. I just will double-check it by EOD.

mike-sul (approved):

EmbeddedAndroid:
Tested in my factory, working as expected.

PR #137, opened 2021-07-12
containers: Include more accurate TUF custom data

doanac:
Inheriting from the LmP build is a little confusing to users as they'll
see a createdAt timestamp in the past and a URI that doesn't point to
the CI job that created this Target.

### DIFF
--- a/apps/target_manager.py
+++ b/apps/target_manager.py
@@ -1,7 +1,8 @@
 # Copyright (c) 2020 Foundries.io
 # SPDX-License-Identifier: Apache-2.0
-
+import datetime
 import logging
+import os
 import json
 from copy import deepcopy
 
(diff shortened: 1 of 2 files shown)


doanac:
@mike-scott FYI

apps/target_manager.py
```diff
@@ -105,6 +106,16 @@ def create_target(targets_json, compose_apps, ota_lite_tag, git_sha, machines, p
             tgt_name = tagmgr.create_target_name(target, version, tag)
             data['targets'][tgt_name] = target
 
+            now = datetime.datetime.utcnow().strftime('%Y-%m-%dT%H:%M:%SZ')
+            target['custom']['createdAt'] = now
```
mike-sul, line 110: iiuc, it is used to be inherited from LmP build?

mike-sul: In this case, we overwrite metadata about an ostree Target which is part of an overall Target.

mike-sul (approved):
I suppose there is no similar issue on the LmP side (customize-target.py)

apps/target_manager.py
doanac: > it is used to be inherited from LmP build?

correct. This sort of makes a linked-list via the origUri to find the actual LmP build now.

> In this case, we overwrite metadata about an ostree Target which is part of an overall Target.

we could preserve the original createdAt field like we do with origUri, but I was worried that's started to get a little verbose and we can deduce the original createdAt timestamp based on the other data we have.

doanac:
> I suppose there is no similar issue on the LmP side (customize-target.py)

meta-updater populates the target in this case. Its influenced by some environment variables. Then our script just sets attributes for compose apps

PR #138, opened 2021-07-14
Fix/use original uri

mike-sul:
There are two fields/URLs in Target's custom that points to a CI build,
`uri` that points to the CI build that generated the given Target,
and `origUri` that points the CI build that generated LmP/rootfs of the given Target.


factory_client.py
```diff
@@ -115,7 +115,7 @@ def get_targets_by_version(self, version):
     def get_target_system_image(self, target: Target, out_dir: str, progress: Progress):
         # https://api.foundries.io/projects/<factory>/lmp/builds/<build-numb>/runs/<machine>/<image-name>-<machine>.wic.gz
 
-        image_base_url = target['custom']['uri']
+        image_base_url = target['custom']['origUri'] if 'origUri' in target['custom'] else target['custom']['uri']
```
doanac, line 118: that's not going to work. if you have:
1. platform build
1. container build
1. container build

The last container build would have 2 for the URI but it really needs to be 1. Maybe we should just back out my change that updates the "uri".

mike-sul: we shouldn't update `origUri` if the `uri` points to a container build? Perhaps, instead if `uri` we need two URI fields one for each sub-target, e.g. `ostreeURI` and `appURI`, the first is updated only by LmP build and just copy `appURI`. Or to make things simpler we could just leave `uri` for LmP and introduce a new field for `appURI`.

doanac: not updating origUri makes sense to me. we just need to add a conditional around:

 https://github.com/foundriesio/ci-scripts/blob/dc5fcba880608651d454986f04f09ccc3bb65595/apps/target_manager.py#L114

to not update that field if its already set.

mike-sul: It will work as long as a LmP build doesn't copy `origUri` from the previous Target, let me check that?

doanac: wouldn't 2 consecutive container builds break the origUri?

mike-sul: not sure I get the last question. LmP build just sets `uri` field, and if it copies `origUri` from a previous Target (N-1) to a new one (N) then during the following container build (N+1), `origUri` is set and it points to N-1.

mike-sul: Yes, it should work, because LmP build doesn't copy nor set `origUri` field, thus if it's set it means that a previous build was a container build.

mike-sul: @doanac Just updated the PR.

doanac (approved):

PR #144, opened 2021-09-02
apps: don't preload and store Apps

mike-sul:
There is no need in dumping and storing of Apps and their images on NFS
storage since their injection into a system image has been moved from
the LmP/bitbake run to separate dedicated run that can dump images by
itself.


doanac (approved):
and just like this, our builds are faster and our storage costs go down.

PR #145, opened 2021-09-09
assemble wic: drop hard-coded partition used for resizing

mike-scott:
Currently the assemble wic script uses a hard-coded value of 2 for resizing the rootfs.

This is not always true.

Let's use the last line of the `fdisk` output as a guide for the last partition (which is the only one that can be resized anyway).

### DIFF
--- a/assemble.py
+++ b/assemble.py
@@ -31,6 +31,7 @@ class WicImage:
     def __init__(self, wic_image_path: str, increase_bytes=None, extra_space=0.2):
         self._path = wic_image_path
         self._mnt_dir = os.path.join('/mnt', 'wic_image_rootfs')
+        self._last_part = 2
         self._resized_image = False
         if increase_bytes:
             self._resize_wic_file(increase_bytes, extra_space)
@@ -47,7 +48,7 @@ class WicImage:
         for line in out.splitlines():
             if self._path in line:
                 self._loop_device = line.split(':', 1)[0]
-                self._wic_device = line.split(':', 1)[0] + 'p2'
+                self._wic_device = line.split(':', 1)[0] + 'p' + str(self._last_part)
                 break
         else:
             raise RuntimeError('Unable to find loop device for wic image')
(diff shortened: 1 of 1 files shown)


mike-scott:
do not merge at the moment still debugging in CI

mike-scott:
ok, debugged and now working for both GPT and MBR partition tables

assemble.py
```diff
@@ -89,7 +89,7 @@ def _resize_wic_file(self, increase_bytes: int, extra_space=0.2):
             'conv=notrunc', 'oflag=append', 'count=' + str(increase_k),
             'seek=' + str(wic_k))
 
-        parted_out = str(subprocess.check_output(['parted', self._path, 'print']))
+        parted_out = subprocess.check_output(['parted', self._path, 'print']).decode('ascii')
         if parted_out.find('Partition Table: gpt') != -1:
```
doanac: str vs an error decoding feels like a weird compromise. my guess we need `.decode("utf-8")` or just `.decode()`. However, we can keep this even simpler and not bother decoding by doing:
```
        parted_out = subprocess.check_output(['parted', self._path, 'print'])
        if parted_out.find(b'Partition Table: gpt') != -1:
```

doanac (commented):
just one nitpick

assemble.py
mike-scott: True, but then I can't use split*() on the byte encoded data.  (Which was the reason behind this change)

doanac: split works for both bytes and strings

mike-scott: This was kind of a nice simplification, last commit changes:
- Drop str() call for `parted_out` which leaves it as byte data
- Adjust `find` and `split` to use to byte data
- Convert final `_last_part` to int for storage -- should break nicely if we every have an issue here

mike-scott: And a few cleanups from that conversion to int.
Re-tested the code path for both GPT and msdos images.

mike-scott:
Added a few cleanups and re-tested against both GPT and msdos partition tables.

mike-sul (approved):
LGTM. I added `lsblk` to the image https://github.com/foundriesio/lmp-images-tools/pull/12
so the script could do something like the following  to identify the "otaroot" partition.

`
bash-5.0# for p in `ls /dev/loop0p*`; do lsblk -no LABEL $p; done
boot
otaroot
`

doanac (approved):

mike-scott:
> LGTM. I added `lsblk` to the image [foundriesio/lmp-images-tools#12](https://github.com/foundriesio/lmp-images-tools/pull/12)
> so the script could do something like the following to identify the "otaroot" partition.
> 
> `bash-5.0# for p in`ls /dev/loop0p*`; do lsblk -no LABEL $p; done boot otaroot `

I don't think the .wic mounted as a loopback when we need to run the resize.. is it?

mike-sul:
> I don't think the .wic mounted as a loopback when we need to run the resize.. is it?

yes, it could be mount, check, umount and then resize. Hey, Mike, the PR is fine and minimalistic , I am just exploring options :).
btw, the `wic` utility  I used originally does the trick without loop back device setup.
```
bash-5.0# /usr/bin/oe/scripts/wic ls /image.wic 
Num     Start        End          Size      Fstype
 1       4194304     47657983     43463680  fat16
 2      50331648    901775359    851443712  ext4
```

PR #153, opened 2021-10-18
containers: Add logic for private container registries

doanac:
Factories may be configured with private container registries. CI
will set these with a "secret" that we can execute to login with.

### DIFF
--- a/apps/login_registries
+++ b/apps/login_registries
@@ -1,0 +1,33 @@
+#!/usr/bin/python3
+
+import json
+from pathlib import Path
+import subprocess
+import sys
+
+from helpers import status
+
+
+if len(sys.argv) != 2:
+    sys.exit("Usage: %s <registry data file>" % sys.argv[0])
+
+with open(sys.argv[1]) as f:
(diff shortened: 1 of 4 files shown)


apps/login_registries
```diff
@@ -0,0 +1,33 @@
...
+        creds_file = Path.home() / ".aws/credentials"
+        creds_file.parent.mkdir()
+        secrets_file = Path("/secrets") / reg["aws_creds_secret_name"]
+        creds_file.write_text(secrets_file.read_text())
+
+        try:
+            cmd = ["aws", "ecr", "get-login-password", "--region", reg["region"]]
+            token = subprocess.check_output(cmd)
+            cmd = ["docker", "login", "--password-stdin", "-u", "AWS", reg["url"]]
+            subprocess.run(cmd, check=True, input=token)
```
mike-sul, line 29: Why this one uses `run` instead of `check_output`?

doanac: if the login fails, i'd like it to show up in stderr/stdout so we can diagnose the issue.

PR #155, opened 2021-11-12
Move to using totally dynamic creds.zip

doanac:
JobServ now sends us the target.sec file. This means there's no need
to rely on the contents of an NFS copy of credentials.zip.

### DIFF
--- a/create-creds
+++ b/create-creds
@@ -4,12 +4,12 @@ import sys
 from helpers import generate_credential_tokens, require_secrets
 
 
-def main(creds_in, creds_out):
-    require_secrets('osftok', 'triggered-by')
-    generate_credential_tokens(creds_in, creds_out)
+def main(creds_out):
+    require_secrets('osftok', 'triggered-by', 'targets.sec')
+    generate_credential_tokens(creds_out)
 
 
 if __name__ == '__main__':
-    if len(sys.argv) != 3:
-        sys.exit('Usage: %s <creds.zip-in> <creds.zip-out>' % sys.argv[0])
-    main(sys.argv[1], sys.argv[2])
+    if len(sys.argv) != 2:
+        sys.exit('Usage: %s <creds.zip-out>' % sys.argv[0])
+    main(sys.argv[1])
--- a/helpers.py
+++ b/helpers.py
@@ -153,20 +153,29 @@ def http_get(url, params=None, **kwargs):
     return response
 
 
-def generate_credential_tokens(creds_zip_in: str, creds_zip_out: str):
-    """Creates an updated creds.zip file that has populated osf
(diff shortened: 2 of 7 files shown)


lmp/bb-config.sh
```diff
@@ -29,10 +29,10 @@ if [ "$ENABLE_PTEST" = "1" ] ; then
     OSTREE_BRANCHNAME="${OSTREE_BRANCHNAME}-ptest"
 fi
 
-if [ -n "$SOTA_PACKED_CREDENTIALS" ] && [ -f $SOTA_PACKED_CREDENTIALS ] ; then
+if [ -f "/secrets/targets.sec" ] ; then
```
mike-sul, line 32: Will it work in the `else` case, I mean if the `targets.sec` secret is not available?

mike-sul:
Just minor comment.
Also, I think I need to learn how JobServ secrets work, not sure I understand how it will work if a worker is running on my dev host for example :).

lmp/bb-config.sh
doanac: yes. this is how people bitbake locally. meta-updater just says something about not being able to publish

mike-sul (approved):

create-creds
```diff
@@ -4,12 +4,12 @@ import sys
 from helpers import generate_credential_tokens, require_secrets
 
 
-def main(creds_in, creds_out):
-    require_secrets('osftok', 'triggered-by')
-    generate_credential_tokens(creds_in, creds_out)
+def main(creds_out):
+    require_secrets('osftok', 'triggered-by', 'targets.sec')
```
mike-sul, line 8: I suppose `root.json` and `targets.pub` are also required secrets.

helpers.py
```diff
@@ -153,20 +153,29 @@ def http_get(url, params=None, **kwargs):
...
-                buf = zin.read(name)
-                if name == 'treehub.json':
-                    data = json.loads(buf.decode())
-                    data['oauth2']['client_id'] = secret('triggered-by')
-                    data['oauth2']['client_secret'] = secret('osftok')
-                    buf = json.dumps(data).encode()
-                zout.writestr(name, buf)
+        zout.writestr('root.json', secret("root.json"))
+        zout.writestr('targets.pub', secret("targets.pub"))
+        zout.writestr('targets.sec', secret("targets.sec"))
```
mike-sul, line 176: I suppose, signing of targets.json requires just a private root role key, this is just garage-sign requires additional elements for some reason?

doanac: yep. its silly but garage sign needs it.

create-creds
doanac: good catch. i'll make that change and just push to master to clean this up.

PR #156, opened 2021-11-23
lmp: Add ssh-agent support for recipes

doanac:
Customer's may have bitbake recipes referencing private ssh+git
repositories. This change allows users to configure Factory support by
naming convention with fioctl-secrets. For example a user can create 2
secrets:

 ssh-github.key - content coming from ssh-keygen (i.e id_rsa)
 ssh-known_hosts - the fingerprint(s) of the server to be connected to

If these are detected, the bitbake build process will launch ssh-agent,
ssh-add the key(s), and set the known_hosts file so that git-clones
won't prompt for user input.

### DIFF
--- a/lmp/bb-build.sh
+++ b/lmp/bb-build.sh
@@ -4,6 +4,21 @@ HERE=$(dirname $(readlink -f $0))
 source $HERE/../helpers.sh
 require_params IMAGE
 
+keys=$(ls /secrets/ssh-*.key 2>/dev/null || true)
+if [ -n "$keys" ] ; then
+	status Found ssh keys, starting an ssh-agent
+	eval `ssh-agent`
+	for x in $keys ; do
+		echo " Adding $x"
+		ssh-add $x
+	done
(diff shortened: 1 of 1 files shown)


doanac:
@mike-scott - i need to set something up in a factory to really test this, but as you can see - the change can be pretty trivial.

doanac:
@mike-scott @mike-sul - this change allows bitbake recipes to reference private repos. 

I hit one bit of friction and created a new fioctl change:

 https://github.com/foundriesio/fioctl/pull/130

The problem is with shells and `fioctl secrets update key="$(cat /file)"`. The shell will truncate the trailing \n in `/file` and ssh-add **requires** that. You can also use the current fioctl with:
~~~
  fioctl secrets update ssh-github.key="$(cat /tmp/ssh-github.key)"$'\n'
~~~

But that's kind of crazy.

mike-sul (approved):
I used to do something like `git -c core.sshCommand="ssh -i private_key_file" clone host:repo.git` , just in case.

PR #157, opened 2021-11-24
lmp/build.sh: publish all imx-boot-machine files

ricardosalveti:
We might have variants of imx-boot (e.g. nohdmi), so make sure
we're publishing all of them together at the root folder.


mike-scott (approved):
LGTM

PR #158, opened 2021-12-03
apps: pass arch list to the publish tool

mike-sul:
- Pass a list of platforms (aka architectures) defined in factory-config.yml::containers::platforms, aka factory architectures) to the compose publish tool, so it can shortlist architectures supported by App's images. Otherwise, CI might publish App that supports non-factory architectures what might lead to subtle issues while running App/containers on a device. 

### DIFF
--- a/apps/apps_publisher.py
+++ b/apps/apps_publisher.py
@@ -16,9 +16,10 @@ logger = logging.getLogger(__name__)
 
 
 class AppsPublisher:
-    def __init__(self, factory, publish_tool: str, registry_host=DockerRegistryClient.DefaultRegistryHost):
+    def __init__(self, factory, publish_tool: str, archs: str, registry_host=DockerRegistryClient.DefaultRegistryHost):
         self._factory = factory
         self._publish_tool = publish_tool
+        self._archs = archs
         self._registry_host = registry_host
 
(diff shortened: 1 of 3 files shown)


mike-sul:
I tested it against a current version of compose-publish as well as against the new one implemented here https://github.com/foundriesio/compose-publish/pull/7 and vice versa. (5 latest builds here https://ci.foundries.io/projects/msul-dev02/lmp/).

apps/apps_publisher.py
```diff
@@ -16,9 +16,10 @@
 
 
 class AppsPublisher:
-    def __init__(self, factory, publish_tool: str, registry_host=DockerRegistryClient.DefaultRegistryHost):
+    def __init__(self, factory, publish_tool: str, archs: [], registry_host=DockerRegistryClient.DefaultRegistryHost):
```
doanac, line 19: i think you are mixing types here. `archs` is actually a string or `cmd_exe` would fail. I think this means the test function at the bottom needs to pass `""` instead of `[]`

mike-sul: yeah, your're right, originally it was actually a list but then I changed it. Updated via forced push.

doanac (approved):

PR #160, opened 2021-12-21
Support the new type of static delta indexes

mike-sul:


lmp/generate-static-deltas
```diff
@@ -89,7 +89,21 @@ def main(creds_zip_file: str, deltas: List[Delta]):
             cmd("ostree", "static-delta", "generate", "--repo=./ostree_repo", "--from", sha, "--to", d.to[0])
             prog.tick()
 
+    # update summary and generate a new type of delta indexes
+    cmd("ostree", "summary", "-u", "--repo=./ostree_repo")
+
+    if os.path.isdir("./ostree_repo/delta-indexes"):
```
doanac, line 95: is this so it works with bold old and lew lmp-image-tools?

mike-sul: Yes.

The only really required change is at #93, it generates "delta-indexes" and updates summary for backward compatibility.

The remaining code is just for debugging/troubleshooting purposes, maybe can be removed in the future.

doanac: could we just get the new lmp-image-tools deployed? its backwards compatible, right? then we wouldn't need the debug at all.

mike-sul: yes, but the changes to ostreehub should be deployed prior to it, since the new  lmp-image-tools uses v2 API and the currently deployed ostreehub v2 API doesn't support a summary update.

doanac: ah. compatibility. i generally hate "TODO comments", but maybe we add one here or create a jira ticket so we remember to remove this?

doanac (approved):

lmp/generate-static-deltas
mike-sul: > maybe we add one here or create a jira ticket so we remember to remove this?
done

doanac (approved):

PR #163, opened 2021-12-23
Determine image preloading type dynamically

mike-sul:
If a user doesn't specify an App type to preload then decide on it taking into account a LmP version.
If LmP > 84 then preload `restorable`  Apps by default. If LmP <= 84 or it's not possible to obtain a LmP version then preload `compose` Apps.
If a user explicitly specifies App type then a LmP version is ignored and the user defined type of Apps is preloaded. 
App type can be defined in factory-config.yml::containers::preloaded_images::app_type.
```
    containers:
      preloaded_images:
        enabled: true
        [app_type: [<restorable | compose | '']>]
```


mike-sul:
Update the doc https://github.com/foundriesio/docs/pull/287 according to this PR as well as changes to repoz https://git.foundries.io/development/cloudplatforms/infrastructure/repoz/-/merge_requests/49.

factory_client.py
```diff
@@ -70,6 +70,32 @@ def _set_apps_commit_hash(self):
         def __getitem__(self, item):
             return self.json[item]
 
+    class Release:
```
doanac, line 73: You can do this with a lot less boiler plate code with something like
```
class Release(NamedTuple):
    lmp_version: int

    @classmethod
    def parse(cls, release_info) -> "Release":
       ...
       return cls(lmp_version)
  
```

factory_client.py
```diff
@@ -143,7 +170,12 @@ def get_target_system_image(self, target: Target, out_dir: str, progress: Progre
         else:
             logger.info('Target system image has been already downloaded: {}'.format(extracted_image_file_path))
 
-        return extracted_image_file_path
+        release_resp = requests.get(os_release_url, headers=self._auth_headers)
+        if release_resp.ok:
+            release_info = dict([line.split('=') for line in release_resp.content.decode().splitlines()])
+        else:
```
doanac, line 176: i can see that for now this will be the normal case, but in the future it seems like we'd be silently ignoring an important error?

factory_client.py
mike-sul: OK, I just usually prefer explicit API, and use "attribute" to explicitly specify public attributes. I think using of dynamic attributes in init class will lead even to lesser boiler plate code in this particular case :). Also, we may use `dataclass` decorator.
Anyway, it's not really important, I'll do it as you suggested.

factory_client.py
mike-sul: > but in the future it seems like we'd be silently ignoring an important error?

I am not sure if it's critical error and a CI build should be stopped in this case.
For the "assemble" run triggered within the LmP job it can be kind of critical error provided that this PR https://github.com/foundriesio/meta-lmp/pull/496 is included into a given LmP version. But, it will take some time for all customers to move to the LmP version that includes `os-release` dumping.
Even if all customers move to the os-release dumping version of LmP they might use `fioctl targets image` to preload Apps into some older version of LmP.
Maybe we can just add an info log informing about the presence or absence of `os-release`?

factory_client.py
mike-sul: I updated the PR according to your suggestion.

factory_client.py
mike-sul: I added an info log if `os-release`  is missing and an error message if `os-release` is present but a parse error happen. Just not sure if we need to (re-)throw exception and exit the CI job/run in this case or just log an error message. On the one hand it's not really critical because the assemble run can do its job without a release info and simply choose a type of Apps to preload based on an explicitly defined type in a user factory (factory-config.yml) and preload `compose` apps if it's not explicitly defined.
Another approach is to exit/fail the CI job if a parse error happens and a type of Apps is not explicitly defined.

doanac (approved):

PR #165, opened 2022-01-04
Remove Apps storing at and reusing via NFS

mike-sul:
Storing and reusing of the "dumped" Apps&images does not bring much benefit, just decreases a running time for some of the preload/assemble CI runs (e.g. LmP build's assemble run if Container's assemble run was executed before).
At the same time, this feature requires usage of NFS, and a required size of it increases as a number of factories and CI builds grows.
Also, it makes an overall implementation more complex.
Thus, removing this feature makes the overall product simpler and more reliable.


mike-sul:
We might consider bringing "assemble" CI runs "closer" to hub.fio/Registry (the same cloud) so image fetching is fast enough to neglect it (GCS becomes our NFS :)).

assemble.py
```diff
@@ -100,24 +99,35 @@ def _resize_wic_file(self, increase_bytes: int, extra_space=0.2):
...
     p = Progress(2, progress)
-    target_app_store = ArchiveTargetAppsStore(apps_archive_dir)
-    target.shortlist = apps_shortlist
-    if not target_app_store.exist(target):
-        logger.info('Container images have not been found, trying to obtain them...')
-        apps_fetcher = TargetAppsFetcher(token, app_preload_dir)
-        apps_fetcher.fetch_target_apps(target, apps_shortlist)
-        apps_fetcher.fetch_apps_images()
-        target_app_store.store(target, apps_fetcher.target_dir(target.name))
+    apps_fetcher = TargetAppsFetcher(token, fetch_dir)
```
doanac, line 105: two new calls to `p.tick()` have been added here. I think line 104 needs to be changed to `p = Progress(4, progress)`

assemble.py
```diff
@@ -158,6 +167,18 @@ def copy_restorable_apps_to_wic(target: FactoryClient.Target, wic_image: str, to
     p.tick()
 
 
+def fetch_apps(target: FactoryClient.Target, token: str, apps_shortlist: list, fetch_dir: str) -> int:
```
doanac, line 170: i'm not seeing what calls this new function?

mike-sul: I am not seeing it neither :), so let's remove it.

doanac (approved):

PR #167, opened 2022-01-05
lmp: Use ssh-agent for repo-sync and OE build

doanac:
Users have both private recipe repositories and private layers. We
have to run two different agents because we repo-sync as root and run
the OE build as "builder".


ricardosalveti (approved):
LGTM

mike-sul (approved):

PR #171, opened 2022-03-01
Fix prebuilt logic for x86 installer images

doanac:


ricardosalveti:
One minor thing, the installer is not specific to x86, we have the same installer image available on generic arm64 targets.

So might be better to just have it as wic-based installer images.

doanac:
@ricardosalveti - is the pattern the same - if rootfs.img exists - you have an installer?

ricardosalveti:
It should be the same, check https://ci.foundries.io/projects/lmp/builds/1455/generic-arm64/artifacts/lmp-base-console-image-generic-arm64.wic.gz.

doanac:
I've removed the x86 references. @mike-sul - whatcha think?

PR #172, opened 2022-03-22
lmp: deploy env and logs even in case of failures

ricardosalveti:
It is quite hard to identify build errors (that are not parsing related)
at the moment, since we don't save the environment and the build log is
only available on successful builds.

Fix by making sure the core files are always stored under ${archive}, as
that folder is always uploaded, even on errors.

### DIFF
--- a/lmp/bb-build.sh
+++ b/lmp/bb-build.sh
@@ -8,6 +8,11 @@ start_ssh_agent
 
 source setup-environment build
 
-bitbake -D ${IMAGE}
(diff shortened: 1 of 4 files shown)


ricardosalveti:
Tested on a successful build, when forcing a parser error and when forcing a build failure, all good.

A lot easier to debug now.

doanac (approved):
I've had dreams of this

mike-sul (approved):

angolini (approved):
LGTM

PR #174, opened 2022-04-05
lmp: Add DEV_MODE variable to be handled on local.conf

angolini:
It allow the variable DEV_MODE to be set as a lmp:params: stanza and
added on the local.conf to be used as needed.

Signed-off-by: Daiane Angolini <daiane.angolini@foundries.io>

### DIFF
--- a/lmp/bb-config.sh
+++ b/lmp/bb-config.sh
@@ -20,6 +20,7 @@ MFGTOOL_FLASH_IMAGE="${MFGTOOL_FLASH_IMAGE-lmp-factory-image}"
 SSTATE_CACHE_MIRROR="${SSTATE_CACHE_MIRROR-/sstate-cache-mirror}"
 USE_FIOTOOLS="${USE_FIOTOOLS-1}"
 OSTREE_API_VERSION="${OSTREE_API_VERSION-v2}"
+DEV_MODE="${DEV_MODE-0}"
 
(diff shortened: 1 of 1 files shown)


lmp/bb-config.sh
```diff
@@ -36,6 +37,12 @@ fi
 
 source setup-environment build
 
+if [ "$DEV_MODE" == "1" ]; then
+	cat << EOFEOF >> conf/local.conf
+DEV_MODE="1"
```
ricardosalveti, line 42: Add space between = to be similar to the over local.conf variables.

angolini: @ricardosalveti done!

lmp/bb-config.sh
```diff
@@ -36,6 +37,12 @@ fi
 
 source setup-environment build
 
+if [ "$DEV_MODE" == "1" ]; then
+	cat << EOFEOF >> conf/local.conf
```
MrCry0, line 41: LGTM.

PS. I'm just curious, why `cat` in-place doc and not `echo`? Like:
```
echo 'DEV_MODE = "1"' >> conf/local.conf
```

angolini: I only copied what is already there. I remember I saw this discussion somewhere, some time ago, but I really don't remember. @MrCry0

ricardosalveti: Yeah, in this case both would work.

I believe we used cat in order to support multiple lines more easily.

PR #176, opened 2022-04-13
fixup! preload: Preserve attributes of preloaded image files

mike-sul:
It turned out that `cp -r` does NOT preserve file/folder attributes.
So, when docker image data are copied with `cp -r` to a wic image, then
all files/folders become owned by `root`. When a host folder
is mounted to an existing container folder then an owner of the folder
within container stays the same and it's changed on a host folder if
they differ.

Using `cp -a` or `cp -p` fixes the issue.


vanmaegima (approved):
LGTM

PR #184, opened 2022-05-20
Sensitive review

angolini:
Depends on https://github.com/foundriesio/meta-lmp/pull/652

I tested it using a factory updated to kirkstone, and with DISABLE_GPLV3=1


lmp/bb-config.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash -e
+
```
ricardosalveti, line 1: I guess this shouldn't be here?

lmp/bb-config.sh
```diff
@@ -163,6 +163,8 @@ if [ "${DISABLE_GPLV3}" = "1" ]; then
 INHERIT += "image-license-checker lmp-disable-gplv3"
 IMAGE_LICENSE_CHECKER_ROOTFS_BLACKLIST = "GPL-3.0 LGPL-3.0 AGPL-3.0"
 IMAGE_LICENSE_CHECKER_NON_ROOTFS_BLACKLIST = "GPL-3.0 LGPL-3.0 AGPL-3.0"
+IMAGE_LICENSE_CHECKER_ROOTFS_DENYLIST = "GPL-3.0* LGPL-3.0* AGPL-3.0*"
+IMAGE_LICENSE_CHECKER_NON_ROOTFS_DENYLIST = "GPL-3.0* LGPL-3.0* AGPL-3.0*"
```
quaresmajose, line 167: The new variables with `DENYLIST` is added but the old ones `BLACKLIST` is not deleted, shouldn't we erase the old ones?

angolini: The ci-script at master is used on all factories, so we let the BLACKLIST for backport compatibility (for the factories using v86 and older)

quaresmajose: Understood, thanks

lmp/bb-config.sh
angolini: ops

thanks for the review @ricardosalveti

ricardosalveti (approved):
LGTM

PR #188, opened 2022-06-08
lmp/build.sh: fix the safe.directory when running the simulator locally

quaresmajose:
== 2022-06-08 13:38:05 Build triggered by change to lmp-manifest
++ pwd
+ manifest='file:///repo/.git -b cfad3167070e3783cac8c6260e9efa63c952e56c'
+ git branch pr-branch cfad3167070e3783cac8c6260e9efa63c952e56c
fatal: unsafe repository ('/repo' is owned by someone else)
To add an exception for this directory, call:

        git config --global --add safe.directory /repo
Script completed with error(s)

### DIFF
--- a/helpers.sh
+++ b/helpers.sh
@@ -49,7 +49,6 @@ function start_ssh_agent {
 function repo_sync {
 	status "Repo syncing sources..."
 
-	git_config
 	if [ -f /secrets/git.http.extraheader ] ; then
 		domain=$(echo $GIT_URL | cut -d/  -f3)
 		status "Adding git config extraheader for $domain"
--- a/lmp/build.sh
+++ b/lmp/build.sh
@@ -7,6 +7,7 @@ require_params MACHINE IMAGE GIT_SHA
 DISTRO="${DISTRO-lmp}"
 
 start_ssh_agent
+git_config
 
 if [[ $GIT_URL == *"/lmp-manifest.git"* ]]; then
 	status "Build triggered by change to lmp-manifest"


ricardosalveti:
What is interesting here is that there is already a similar config (global) set when repo_sync gets called.

https://github.com/foundriesio/ci-scripts/blob/master/lmp/build.sh#L20 -> https://github.com/foundriesio/ci-scripts/blob/master/helpers.sh#L52 -> https://github.com/foundriesio/ci-scripts/blob/master/helpers.sh#L20, so in theory that is already set and with the same user.

quaresmajose:
> What is interesting here is that there is already a similar config (global) set when repo_sync gets called.

the repo_sync is called after the `git branch pr-branch $GIT_SHA ` that is the first git call inside the /repo, if we can call the `git_config` from the helper on startup of the `build.sh` it solves the issue as well

ricardosalveti:
Got it, so the issue happens because there is a git manipulation before the config is set. And why is that only being reproduced when running with simulator?

quaresmajose:
@doanac  told me `my guess is that jobserv executes the simulator as root and that you didn't run the script as root` and locally we run it as a regular user

quaresmajose:
> Yeah, I think calling it when starting build.sh would be better as we can just leverage the helper function.

Tomorrow I update

MrCry0:
@quaresmajose in the commit message: `s/firs/first/`

quaresmajose:
thanks @MrCry0

MrCry0 (approved):
LGTM

ricardosalveti (approved):
LGTM

PR #189, opened 2022-06-09
lmp: add BITBAKE_EXTRA_ARGS to use additional the bitbake arguments

quaresmajose:
This will add the option to change the default bitbake arguments using the variable `BITBAKE_EXTRA_ARGS` 

NOTE: currently in the version we have it is already possible to change the default with:
```
ci_scripts:
    url: https://github.com/quaresmajose/ci-scripts
    git_ref: wip
```

### DIFF
--- a/lmp/bb-build.sh
+++ b/lmp/bb-build.sh
@@ -18,4 +18,4 @@ bitbake -e ${IMAGE} > ${archive}/bitbake_image_env.txt
 # Setscene (cache), failures not critical
 bitbake --setscene-only ${IMAGE} || true
 
-bitbake -D ${IMAGE}
+bitbake -D ${BITBAKE_EXTRA_ARGS} ${IMAGE}


lmp/bb-build.sh
```diff
@@ -18,4 +18,4 @@ bitbake -e ${IMAGE} > ${archive}/bitbake_image_env.txt
 # Setscene (cache), failures not critical
 bitbake --setscene-only ${IMAGE} || true
 
-bitbake -D ${IMAGE}
+bitbake ${BITBAKE_ARGS} ${IMAGE}
```
angolini, line 21: removing -D by default will impact in the debug artifacts we provide on each factory. I don't know if I like it, but I don't have hard feelings, so I'm ok with the change.

However, I would prefer to have `BITBAKE_ARGS` to have `-D` as initial default, at least.

angolini (approved):
LGTM

lmp/bb-build.sh
MrCry0: @angolini Line 26 in `lmp/bb-config.sh` does exactly what you ask.

MrCry0 (approved):

lmp/bb-build.sh
ricardosalveti: Right, but do we want to disable -D here? I know this is an advanced customization, but I wonder if it would be better to force -D instead and allow extra arguments via BITBAKE_ARGS.

@quaresmajose how were you planning on using this?

ricardosalveti: Yup, it is fine to have it as default when not set, I'm just wondering if there is any bad side effect of forcing -D on all cases (and allowing additional arguments).

quaresmajose: I think no, it can work like you said: `-D` forced and `BITBAKE_ARGS` to extend

ricardosalveti: Cool, then lets keep -D and just allow extra args instead (that will help keeping the builds somehow behaving similarly).

ricardosalveti (approved):
LGTM

Issue #148, opened 2021-09-28
EXTRA_TAGS broke after move to buildx


doanac:
https://github.com/foundriesio/ci-scripts/blob/cb79078d4903278f22cef6b7f1c6db11ac2750a1/apps/build.sh#L212

It looks like the EXTRA_TAGS option may have been broken with the move to docker-buildx.

doanac:
@munoz0raul - FYI

PR #198, opened 2022-07-06
apps: Clean up logic around multi arch manifests

doanac:
We've had a buggy way of creating multi-arch manifests since the
inception of this tool. This predated docker-manifest and the ability of
CI builds to access artifacts from another run. The issue is that we
build each platform concurrently. Therefore, we don't know which run
is going to complete last and be able to create the multi-arch manifest.

This lead to the ugly hack where we basically did:

 docker manifest create && docker manifest push || true

Only the last run will actually publish the manifest. However, you
periodically have infra issues where the manifest push operation
actually fails. You don't *see* that and we move on the publish run
which will eventually fail to pin this multi-arch reference since it
doesn't exist. Then someone in support has to back-track to find the
source of the issue was one of the other CI Runs.

This change fixes the issue by archiving bits of what
docker-manifest-create does for each CI run. The *publish* run can then
combine them all and *publish* the multi-arch manifests. This has a big
improvement from the previous approach where we created the manifest by
*tags*. This change does by sha256 making it impossible to tamper one of
the tags between the build and publish actions.

This work was almost impossible to do without also fixing:

 https://github.com/foundriesio/ci-scripts/issues/148

That was a long-standing issue broken from the move from the old
"mainfest-tool" and to buildx+ docker-manifest. It has a nice
side-effect in that we no longer need to define "MANIFEST_PLATFORMS" as
we've documented:

 https://docs.foundries.io/latest/reference-manual/docker/containers.html#examples

it was left-over junk from manifest-tool.

### DIFF
--- a/apps/build.sh
+++ b/apps/build.sh
@@ -196,26 +196,24 @@ for x in $IMAGES ; do
 			run docker push ${ct_base}:$TAG-$ARCH
 		fi
 
+		run docker manifest create ${ct_base}:${H_BUILD}_$TAG ${ct_base}:$TAG-$ARCH
+		run docker manifest create ${ct_base}:${LATEST} ${ct_base}:$TAG-$ARCH
+
 		var="EXTRA_TAGS_$ARCH"
 		for t in $(eval echo "\$$var") ; do
-			status "Tagging and pushing extra tag defined in docker-build.conf: $t"
-			run docker tag ${ct_base}:$TAG-$ARCH ${ct_base}:$TAG-$t
-			run docker push ${ct_base}:$TAG-$t
+			status "Handling manifest logic for $var"
+
+			tmp=$HOME/.docker/manifests/hub.foundries.io_${FACTORY}_${x}-${H_BUILD}_${TAG}
+			cp ${tmp}/hub.foundries.io_${FACTORY}_${x}-${TAG}-${ARCH} ${tmp}/hub.foundries.io_${FACTORY}_${x}-${TAG}-${t}
+			run docker manifest annotate ${ct_base}:${H_BUILD}_${TAG} ${ct_base}:${TAG}-$t --arch $t
+
+			tmp=$HOME/.docker/manifests/hub.foundries.io_${FACTORY}_${x}-${LATEST}
+			cp ${tmp}/hub.foundries.io_${FACTORY}_${x}-${TAG}-${ARCH} ${tmp}/hub.foundries.io_${FACTORY}_${x}-${TAG}-${t}
+			run docker manifest annotate ${ct_base}:${LATEST} ${ct_base}:${TAG}-$t --arch $t
 		done
 
 		echo "Build step $((completed+2)) of $total is complete"
 
-		# Convert the old manifest-tool formatted arguments of:
-		#  linux/amd64,linux/arm,linux/arm64
-		# into amd64 arm arm64
-		manifest_args=""
-		for arch in `echo $MANIFEST_PLATFORMS | sed -e 's/linux\///g' -e 's/,/ /g'` ; do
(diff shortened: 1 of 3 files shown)


apps/build.sh
```diff
@@ -196,26 +196,24 @@ for x in $IMAGES ; do
 			run docker push ${ct_base}:$TAG-$ARCH
 		fi
 
+		run docker manifest create ${ct_base}:${H_BUILD}_$TAG ${ct_base}:$TAG-$ARCH
+		run docker manifest create ${ct_base}:${LATEST} ${ct_base}:$TAG-$ARCH
+
 		var="EXTRA_TAGS_$ARCH"
 		for t in $(eval echo "\$$var") ; do
 			status "Tagging and pushing extra tag defined in docker-build.conf: $t"
```
mike-sul, line 204: Maybe the comment should be adjusted accordingly?

apps/build.sh
```diff
@@ -196,26 +196,24 @@ for x in $IMAGES ; do
...
-			run docker tag ${ct_base}:$TAG-$ARCH ${ct_base}:$TAG-$t
-			run docker push ${ct_base}:$TAG-$t
+
+			tmp=$HOME/.docker/manifests/hub.foundries.io_${FACTORY}_${x}-${H_BUILD}_${TAG}
+			cp ${tmp}/hub.foundries.io_${FACTORY}_${x}-${TAG}-${ARCH} ${tmp}/hub.foundries.io_${FACTORY}_${x}-${TAG}-${t}
+			run docker manifest annotate ${ct_base}:${H_BUILD}_${TAG} ${ct_base}:${TAG}-$t --arch $t
+
+			tmp=$HOME/.docker/manifests/hub.foundries.io_${FACTORY}_${x}-${LATEST}
```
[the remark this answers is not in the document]
doanac: line 237 of this file below.

apps/build.sh
```diff
@@ -196,26 +196,24 @@ for x in $IMAGES ; do
 			run docker push ${ct_base}:$TAG-$ARCH
 		fi
 
+		run docker manifest create ${ct_base}:${H_BUILD}_$TAG ${ct_base}:$TAG-$ARCH
+		run docker manifest create ${ct_base}:${LATEST} ${ct_base}:$TAG-$ARCH
+
 		var="EXTRA_TAGS_$ARCH"
 		for t in $(eval echo "\$$var") ; do
```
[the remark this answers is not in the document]
doanac: It was just a lot easier to do it here since this logic of sourcing docker-build.conf correctly is done here. Otherwise we have to copy that code there and most of publish is done as python.

apps/publish_manifest_lists.py
```diff
@@ -0,0 +1,53 @@
...
+                    pass
+                with open(path, "w") as f:
+                    json.dump(mf, f)
+
+    for tag in tags.keys():
+        mfdir = os.path.join(manifests_dir, tag)
+        names = os.listdir(mfdir)
+        tag = tag.replace(f"_{factory}_", f"/{factory}/")
+        if mfdir.endswith(latest_tag):
+            tag = tag.replace(f"-{latest_tag}", f":{latest_tag}")
```
mike-sul, line 49: It would be great to add a comment containing a tag format this logic composes.

apps/build.sh
doanac: https://github.com/foundriesio/ci-scripts/pull/198/commits/99c11643beccaf59445136efd38d6ecdaf8d09f8

apps/publish_manifest_lists.py
doanac: https://github.com/foundriesio/ci-scripts/pull/198/commits/7ca2bc845b23d95f89cd604c5c3022b5e2bad22c

mike-sul (approved):

PR #199, opened 2022-07-06
preload: Check if a registry creds file exists

mike-sul:
- Make sure that the `registry creds` secret file exists before loading
  it into a dict.

### DIFF
--- a/preload_apps.py
+++ b/preload_apps.py
@@ -70,7 +70,12 @@ if __name__ == '__main__':
             target_json = json.load(target_json_file)
 
         registry_creds = []
-        if params.registry_creds_file:
+        # If a file that contains an info about 3rd party registry credentials exists, then load it
+        # so the follow-up logic logins into the registries.
+        # If the file exists and file opening or reading fails then no point to continue since the fact that
+        # the creds file exists implies a need in login into registries and the file reading becomes mandatory.
+        # If the file does not exist, then it's considered as a normal flow - no images in 3rd party registries.
+        if params.registry_creds_file and os.path.exists(params.registry_creds_file):
             with open(params.registry_creds_file) as f:
                 registry_creds = json.load(f)
 


preload_apps.py
```diff
@@ -70,7 +70,7 @@ def get_args():
             target_json = json.load(target_json_file)
 
         registry_creds = []
-        if params.registry_creds_file:
+        if params.registry_creds_file and os.path.exists(params.registry_creds_file):
```
MrCry0, line 73: open() raises an exception if a file does not exist or can't be opened. Wouldn't it be better to handle this exception?

MrCry0: So, say, if the file exists and can't be opened, the script should fail?

doanac: i think the better way to handle possible file not existing is to always just do:
```
try:
    with open(params.registry_creds_file) as f:
        ...
except FileNotFoundError:
    # This is okay
    pass
```
Then it handles @MrCry0's comment and is also more "fool-proof" in more generic settings where a file could disappear between checking for it existing and reading it.

doanac: correct - and failing to *read* will give a different exception. the only foolproof way to try and read and see if a file doesn't exist is to catch an exception. I know it's an edge is this case - but it's just the canonical way to read things safely and catch the not-exists case.

MrCry0: @mike-sul Now I see, your way is the best for such logic.

mike-sul: >where a file could disappear between checking for it existing and reading it.

Again, if this file is defined and it magically disappear then it means an error situation, no point to continue since it will fails while trying to fetch images from 3rd party registries without auth.

doanac: my point is this is generally how *not* to read a file correctly. my suggestion *always* works. This method works under this condition, so every time you do this type approach, you get people asking questions in a review and you have to explain yourself.

mike-sul: Well, I've got different opinion on it. I don't believe in "general" solution, if this is the only correct way to read a file why it's not embedded into `open` call :). Quite often if a file is missing then it's an error/exception so that `try catch` is not needed at all.

doanac: it just seems redundant. `exits` makes a `stat` to see if the file exists and `open` makes an `open` syscall and checks if the open failed because the file didn't exist.

but - you've argued for this passionately several times now. there's no point changing it.

mike-scott (approved):
LGTM

preload_apps.py
MrCry0: @mike-sul Would you mind adding your comment about the logic in the code? It helps a lot in the future.

doanac (approved):

preload_apps.py
mike-sul: @MrCry0 I added the comment.

mike-sul: Speaking of performance - https://therenegadecoder.com/code/how-to-check-if-a-file-exists-in-python/#performance.

PR #201, opened 2022-07-14
apps: Fix flaky issue with buildkit

doanac:
We have a customer using a private Google Artifact Registry instance
that's hit buildkit problems. Moving from 0.8.3 to 0.10.3 resolved them
at first. However, as they've added more containers to their factory, we
are seeing a need to create a new context to make them all build
correctly.

This change creates a context for each container and also makes it
easier for us to change what version of buildkit a customer uses.

### DIFF
--- a/apps/build.sh
+++ b/apps/build.sh
@@ -16,6 +16,8 @@ HERE=$(dirname $(readlink -f $0))
 
 require_params FACTORY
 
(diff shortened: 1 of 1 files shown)


doanac:
this is running in a customer factory now. I'm thinking i'll deploy on monday to let things settle down more from today's excited.

mike-sul:
LGTM

PR #204, opened 2022-07-29
apps: Use the golang compose CLI tool

mike-sul:

### DIFF
--- a/apps/compose_apps.py
+++ b/apps/compose_apps.py
@@ -16,7 +16,7 @@ class ComposeApps:
     DisabledSuffix = '.disabled'
 
     class App:
-        DockerComposeTool = 'docker-compose'
+        DockerComposeTool = 'docker'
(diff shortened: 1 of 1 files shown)


doanac (approved):
is there an easy way to test this in our factories before pushing to everyone?

mike-sul:
> is there an easy way to test this in our factories before pushing to everyone?

Isn't better if it fails at this stage then later on a device which already running `docker compose` anyway?

mike-sul:
I'll re-run the unit tests once the new image (lmp-image-tools) is built

mike-sul:
I can enable it my factory just to double check.

doanac:
> Isn't better if it fails at this stage then later on a device which already running docker compose anyway?

Wasn't talking about devices. I meant - we can test it in CI a bit to make sure we aren't breaking customer ci unexpectedly. looks like its easy to test it, so no big concern

PR #206, opened 2022-08-02
preload: Fallback to `docker-compose` if no `docker compose`

mike-sul:

### DIFF
--- a/apps/compose_apps.py
+++ b/apps/compose_apps.py
@@ -38,7 +38,10 @@ class ComposeApps:
             self._image_downloader_cls = image_downloader_cls
 
             args = [self.DockerComposeTool, 'compose', '-f', self.ComposeFile, 'config']
-            cmd_exe(*args, cwd=self.dir)
+            try:
+                cmd_exe(*args, cwd=self.dir)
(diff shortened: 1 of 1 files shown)


apps/compose_apps.py
```diff
@@ -38,7 +38,10 @@ def __init__(self, name, app_dir, image_downloader_cls=DockerDownloader):
             self._image_downloader_cls = image_downloader_cls
 
             args = [self.DockerComposeTool, 'compose', '-f', self.ComposeFile, 'config']
-            cmd_exe(*args, cwd=self.dir)
+            try:
+                cmd_exe(*args, cwd=self.dir)
+            except FileNotFoundError:
```
doanac, line 43: Might be worth adding a comment here why we have to have this. I'm guess it's because oe-builtin preload might be on an older LmP that doesn't have the new compose?

mike-sul: > I'm guess it's because oe-builtin preload might be on an older LmP that doesn't have the new compose?

Yes. This is a temporal fix.
I am going to add the golang compose to lmp-sdk.

PR #208, opened 2022-08-08
LmP SBOMs

doanac:
These changes result in OE and Container builds both producing SBOMs in a consistent location so that we can start to build tooling to share them with users.

### DIFF
--- a/apps/build.sh
+++ b/apps/build.sh
@@ -223,6 +223,11 @@ for x in $IMAGES ; do
 		echo "osftoken not provided, skipping publishing step"
 	fi
 
+	status "Doing a syft SBOM scan"
+	sbom_dst=/archive/sboms/${ct_base}/${ARCH}.sdpx.json
+	mkdir -p $(dirname $sbom_dst)
+	syft ${ct_base}:$TAG-$ARCH -o spdx-json > $sbom_dst
+
 	if [ -n "$TEST_CMD" ] ; then
 		status Running test command inside container: $TEST_CMD
 		echo "<testcase name=\"test-$x\">" >> /archive/junit.xml
@@ -241,3 +246,5 @@ done
 # Store the manifest so we can use them in the publish run. A brand new
 # factory may not have built any containers, so ensure the directory exists
 [ -d $HOME/.docker/manifests ] && mv $HOME/.docker/manifests /archive/manifests || echo 'no manifests to archive'
+
+PYTHONPATH=${HERE}/.. python3 ${HERE}/generate_non_factory_sboms.py --arch=$ARCH
--- a/apps/compose_apps.py
+++ b/apps/compose_apps.py
@@ -5,6 +5,8 @@ import logging
 import os
 import yaml
 
+from expandvars import expandvars
+
 from helpers import cmd as cmd_exe
 from apps.image_downloader import DockerDownloader
 
@@ -27,7 +29,7 @@ class ComposeApps:
                 raise ValueError('docker-compose.yaml file found. This must be named docker-compose.yml')
             return exists
 
-        def __init__(self, name, app_dir, image_downloader_cls=DockerDownloader):
+        def __init__(self, name, app_dir, image_downloader_cls=DockerDownloader, quiet=False):
             if not self.is_compose_app_dir(app_dir):
                 raise Exception('Compose App dir {} does not contain a compose file {}'
                                 .format(app_dir, self.ComposeFile))
@@ -38,6 +40,8 @@ class ComposeApps:
             self._image_downloader_cls = image_downloader_cls
 
             args = [self.DockerComposeTool, 'compose', '-f', self.ComposeFile, 'config']
+            if quiet:
+                args.append('--quiet')
             cmd_exe(*args, cwd=self.dir)
 
             with open(self.file) as compose_file:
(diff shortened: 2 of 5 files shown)


doanac:
@mike-sul @ricardosalveti - this is still a little bit "work in progress". I'd like to work on the ota-lite api to make sure this approach works okay, but I don't foresee much, if anything, changing with this. So - now would be a good time to say this is a good/bad approach

ricardosalveti (approved):
LGTM

ricardosalveti:
Having under sbom sounds good and it is also a fixed location.

apps/build.sh
```diff
@@ -223,6 +223,11 @@ for x in $IMAGES ; do
 		echo "osftoken not provided, skipping publishing step"
 	fi
 
+	status "Doing a syft SBOM scan"
+	sbom_dst=/archive/sboms/${ct_base}/${ARCH}.sdpx.json
+	mkdir -p $(dirname $sbom_dst)
+	syft ${ct_base}:$TAG-$ARCH -o spdx-json > $sbom_dst
```
mike-sul, line 229: How long does it take to scan an image? If it takes some time I wonder if it's possible to output some progress.

apps/compose_apps.py
```diff
@@ -37,8 +37,9 @@ def __init__(self, name, app_dir, image_downloader_cls=DockerDownloader):
 
             self._image_downloader_cls = image_downloader_cls
 
-            args = [self.DockerComposeTool, 'compose', '-f', self.ComposeFile, 'config']
-            cmd_exe(*args, cwd=self.dir)
+            if self.DockerComposeTool:
```
mike-sul, line 40: Is it temporal change or is it done for some purpose?

mike-sul: I see now why you did it.

apps/build.sh
doanac: it sends progress to stderr

apps/generate_non_factory_sboms.py
```diff
@@ -0,0 +1,57 @@
...
+
+    # We don't need to validate the compose content here - just need to
+    # iterate service images.
+    ComposeApps.App.DockerComposeTool = None
+
+    apps = ComposeApps(args.apps_root)
+    analyzed = {}
+    for app in apps:
+        for img in app.images():
+            img = expandvars(img)
```
mike-sul, line 24: Maybe it makes sense to move it to `ComposeApps.Apps.images`.

apps/generate_non_factory_sboms.py
```diff
@@ -0,0 +1,57 @@
...
+from helpers import status
+
+
+def main(args):
+    status("Loading docker-compose files to find non-factory containers")
+    factory_prefix = f"hub.foundries.io/{args.factory}/"
+
+    # We don't need to validate the compose content here - just need to
+    # iterate service images.
+    ComposeApps.App.DockerComposeTool = None
```
mike-sul, line 18: Is validation causes any issues or there was another reason why you turned it off?

doanac: it gets noisy when you see the CI output. it starts dumping the compose content to stdout and doesn't make much sense why

mike-sul:
I wonder whether it makes sense to do scanning for both hub.f.io and 3rd party hosted images within the same script. It looks like implementation will be encapsulated in one spot in this case.

apps/generate_non_factory_sboms.py
mike-sul: Probably we need to add `--quite` flag` to the `config` cmd.

doanac:
> I wonder whether it makes sense to do scanning for both hub.f.io and 3rd party hosted images within the same script. It looks like implementation will be encapsulated in one spot in this case.

I thought about that and actually implemented that. However, it's inefficient. The publish script has to download all the per-arch container images that we'd built earlier.

apps/generate_non_factory_sboms.py
doanac: https://github.com/foundriesio/ci-scripts/pull/208/commits/892c7a21f8571d4d1b523dda05f7d88502787468

apps/generate_non_factory_sboms.py
doanac: https://github.com/foundriesio/ci-scripts/pull/208/commits/19c68c605664807fc8dca2905a2752d8eb8d3657

doanac:
@mike-sul - good thing we are re-testing this. the move to `docker compose` has broken my change. I need to pull golang compose into https://github.com/foundriesio/docker-dind

apps/compose_apps.py
```diff
@@ -38,17 +41,23 @@ def __init__(self, name, app_dir, image_downloader_cls=DockerDownloader):
             self._image_downloader_cls = image_downloader_cls
 
             args = [self.DockerComposeTool, 'compose', '-f', self.ComposeFile, 'config']
-            cmd_exe(*args, cwd=self.dir)
+            if quiet:
```
mike-sul, line 44: I think that's fine. I just thought about another option

```
if quite:
  args.append('--quite')

```

mike-sul: > The publish script has to download all the per-arch container images

I thought about it too. It's just not necessary, just thought, we know a tag of images we build, i.e. < commit short hash>.

mike-sul:
> I need to pull golang compose into https://github.com/foundriesio/docker-dind

mike-sul:
@doanac I just closed this PR by mistake, I reopened it.

apps/generate_non_factory_sboms.py
doanac: i've done another force push to use your suggestion and have a new ci container. everything should be in place now

mike-sul (approved):

PR #210, opened 2022-08-19
lmp/jobserv: run bitbake with --continue for github_pr

quaresmajose:
Continue as much as possible after an error. While the
target that failed and anything depending on it cannot
be built, as much as possible will be built before
stopping.

### DIFF
--- a/lmp/jobserv.yml
+++ b/lmp/jobserv.yml
@@ -276,6 +276,7 @@ triggers:
         triggers:
           - name: ptest-build-{loop}
         params:
+          BITBAKE_EXTRA_ARGS: --continue
           IMAGE: lmp-base-console-image
         script-repo:
           name: fio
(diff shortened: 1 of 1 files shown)


doanac (commented):
this is really interesting. would this be something we should just do by default for all factories?

quaresmajose:
> this is really interesting. would this be something we should just do by default for all factories?

it helps if we can't test all the possible combinations of builds locally as it will helps shows how many broken things we have.
in the factories we usually have only one target machine just like a single image so when the user push their changes they have already been tested locally.

also to mention this will increase significantly the overhead cpu/memory/io for each build.

ricardosalveti:
I think it is better to have it enabled here only because we do want to fail fast on subscribers (as we expect the errors to be isolated an probably caused by the user).

ricardosalveti (approved):
LGTM

PR #212, opened 2022-08-22
lmp/bb-config: accepte the synaptics-killswitch license on raspberrypi

quaresmajose:
fixes the following issue:

```
| ERROR: Nothing RPROVIDES 'linux-firmware-rpidistro-bcm43455' (but /srv/oe/build/conf/../../layers/meta-lmp/meta-lmp-base/recipes-samples/images/lmp-base-console-image.bb, /srv/oe/build/conf/../../layers/openembedded-core/meta/recipes-core/packagegroups/packagegroup-base.bb RDEPENDS on or otherwise requires it)
| linux-firmware-rpidistro RPROVIDES linux-firmware-rpidistro-bcm43455 but was skipped: because it has a restricted license 'synaptics-killswitch'. Which is not listed in LICENSE_FLAGS_ACCEPTED
| ERROR: Required build target 'lmp-base-console-image' has no buildable providers.
```

### DIFF
--- a/lmp/bb-config.sh
+++ b/lmp/bb-config.sh
@@ -65,6 +65,7 @@ ACCEPT_EULA:stm32mp1-disco = "1"
 ACCEPT_EULA:stm32mp1-eval = "1"
 ACCEPT_EULA:stm32mp15-disco = "1"
 ACCEPT_EULA:stm32mp15-eval = "1"
+LICENSE_FLAGS_ACCEPTED:append:rpi = " synaptics-killswitch"
 EOFEOF
 fi
 


lmp/bb-config.sh
```diff
@@ -72,6 +72,7 @@ cat << EOFEOF >> conf/local.conf
 CONNECTIVITY_CHECK_URIS = "https://www.google.com/"
 
 ACCEPT_FSL_EULA = "1"
+LICENSE_FLAGS_ACCEPTED:rpi += "synaptics-killswitch"
```
ricardosalveti, line 75: Maybe use append instead? Otherwise if LICENSE_FLAGS_ACCEPTED is set to a generic value, it will be replaced entirely due the :rpi override (unless it started with LICENSE_FLAGS_ACCEPTED:rpi = "foobar").

quaresmajose: good catch! thanks

angolini (approved):

lmp/bb-config.sh
```diff
@@ -72,6 +72,7 @@ cat << EOFEOF >> conf/local.conf
 CONNECTIVITY_CHECK_URIS = "https://www.google.com/"
 
 ACCEPT_FSL_EULA = "1"
+LICENSE_FLAGS_ACCEPTED:append:rpi = " synaptics-killswitch"
```
ricardosalveti, line 75: On other thing I just remembered, this change will be used on every factory, and some might still not be using the new syntax, so we need to do something like done for other variables, like:

```
if [ "$CONF_VERSION" == "2" ]; then
        cat << EOFEOF >> conf/local.conf
LICENSE_FLAGS_ACCEPTED:append:rpi = " synaptics-killswitch"
EOFEOF
fi
```

Or just reuse one of the if/else blocks that are already there.

ricardosalveti: And it is fine to only cover the new syntax as we know this is not an issue on older meta-rpi layers.

quaresmajose: > On other thing I just remembered, this change will be used on every factory, and some might still not be using the new syntax, so we need to do something like done for other variables, like:
> 
> ```
> if [ "$CONF_VERSION" == "2" ]; then
>         cat << EOFEOF >> conf/local.conf
> LICENSE_FLAGS_ACCEPTED:append:rpi = " synaptics-killswitch"
> EOFEOF
> fi
> ```
> 
> Or just reuse one of the if/else blocks that are already there.

I had already thought about suggesting deleting the old syntax but now I understand why it still exists.

ricardosalveti (approved):
LGTM

PR #214, opened 2022-09-15
assemble.py: Fix sporadic losetup issue

doanac:
We've been fighting losetup for a while thinking it was some issue with the WIC file not being flushed to disk. Turns out the ENOTEXIST error was actually from a missing /dev/loopX node. This shuffles things around to make the dirty bits more isolated.

### DIFF
--- a/assemble.py
+++ b/assemble.py
@@ -19,6 +19,50 @@ from factory_client import FactoryClient
 logger = logging.getLogger("System Image Assembler")
 
 
+def remount_dev():
+    # containers don't see changes to /dev, so we have to hack around
+    # this by basically mounting a new /dev. The idea was inspired by
+    # this comment:
+    #  https://github.com/moby/moby/issues/27886#issuecomment-257244027
+    if getattr(remount_dev, 'called', None):
+        # you can't call remount_dev dev a 2nd time without a umount, but
+        # it also sometimes *fails* in between doing multiple target assemblies.
+        # My guess is its something with dockerd getting spawned and stopped.
+        # So just *try* to umount and hope for the best.
+        subprocess.call(["umount", "/dev"])
+    cmd('mount', '-t', 'devtmpfs', 'devtmpfs', '/dev')
+    setattr(remount_dev, 'called', True)
+
+
+def losetup(path: str) -> str:
+    # losetup can be tricky to run in a container. Due to the issues with
+    # /dev noted about in `remount_dev`, we periodically will have losetup
+    # run like:
+    #   bash-5.1# losetup -f /file
+    #   losetup: /file: No such file or directory
+    # However, losetup is giving misleading error message. strace will show the
+    # real issue is that losetup creates new loop device, say loop2. However,
+    # there's some kind of timing issue where /dev/loop2 doesn't always show
+    # up, thus the "No such file or directory". When this happens, we need to
+    # remount /dev in order to see the new device.
+    try:
+        cmd('losetup', '-P', '-f', path)
+    except subprocess.CalledProcessError:
+        logger.error('losetup bug found, remounting /dev to work around')
+        remount_dev()
+        cmd('losetup', '-P', '-f', path)
+
+    # The -P in losetup scans for partitions and will create entries like:
+    # /dev/loopXp1. Since these are new /dev entries, we have to remount /dev
+    remount_dev()
+
+    out = cmd('losetup', '-a', capture=True).decode()
+    for line in out.splitlines():
+        if path in line:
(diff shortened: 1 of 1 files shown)


doanac:
gotta love it - as soon as i created the fix, my laptop could not longer reproduce the issue.

doanac:
Okay - this is actually *working* a bit better. I just re-ran a customer CI job:
```
|--
=$  losetup -P -f /archive/lmp-factory-image-imx8mm-lpddr4-evk.wic
| losetup: /archive/lmp-factory-image-imx8mm-lpddr4-evk.wic: No such file or directory
2022-09-15 03:10:20,096 ERROR: assemble: losetup bug found, remounting /dev to work around
|--
=$  mount -t devtmpfs devtmpfs /dev
|--
=$  losetup -P -f /archive/lmp-factory-image-imx8mm-lpddr4-evk.wic
|--
=$  umount /dev
|--
=$  mount -t devtmpfs devtmpfs /dev
|--
=$  losetup -a
| /dev/loop0: 0 /var/lib/snapd/snaps/amazon-ssm-agent_5656.snap
| /dev/loop1: 0 /var/lib/snapd/snaps/core18_2409.snap
| /dev/loop2: 0 /var/lib/snapd/snaps/lxd_22753.snap
| /dev/loop3: 0 /var/lib/snapd/snaps/core20_1518.snap
| /dev/loop4: 0 /var/lib/snapd/snaps/snapd_16292.snap
| /dev/loop5: 0 /var/lib/snapd/snaps/snapd_16778.snap
| /dev/loop6: 0 /var/lib/snapd/snaps/core18_2566.snap
| /dev/loop7: 0 /var/lib/snapd/snaps/core20_1623.snap
| /dev/loop8: 0 /archive/lmp-factory-image-imx8mm-lpddr4-evk.wic
| /dev/loop9: 0 /archive/lmp-factory-image-imx8mm-lpddr4-evk.wic
|--
=$  e2fsck -y -f /dev/loop8p2
```
I'm seeing a failure further down with a call to umount. I *think* maybe i can ignore the umount failure and this might work. Trying that now.

doanac:
its working now. had to make a couple of other small tweaks

mike-scott (approved):
I really like that we are removing the "try 6 times" brute force approach.  LGTM.

mike-sul:
That's great! Thanks for looking into it.

BTW, I noticed that it is 100% reproducible just after reboot on my laptop. Also, it happens on a "fresh" VM that the scaler has just spun up.

assemble.py
```diff
@@ -19,6 +19,50 @@
...
+def losetup(path: str) -> str:
+    # losetup can be tricky to run in a container. Due to the issues with
+    # /dev noted about in `remount_dev`, we periodically will have losetup
+    # run like:
+    #   bash-5.1# losetup -f /file
+    #   losetup: /file: No such file or directory
+    # However, losetup is giving misleading error message. strace will show the
+    # real issue is that losetup creates new loop device, say loop2. However,
+    # there's some kind of timing issue where /dev/loop2 doesn't always show
+    # up, thus the "No sush file or directory". When this happens, we need to
```
MrCry0, line 46: s/sush/such/

assemble.py
```diff
@@ -19,6 +19,50 @@
...
+    # remount /dev in order to see the new device.
+    try:
+        cmd('losetup', '-P', '-f', path)
```
quaresmajose, line 49: `kpartx -a -v path` can do the same job and maybe they don't have the same timing issue of losetup.
https://linux.die.net/man/8/kpartx
I don't do any test and my comment is only a suggestion.

mike-sul: I am not sure this is related to timing, it rather has to do with dockerd/containerd ability to propagate changes from a host to container. If `losetup` is invoked in container changes has to made on a host to create a new lo device, and then this new device should become visible within container (that's my understanding).

quaresmajose: understand, so maybe `kpartx` will suffer from the same problem of the `losetup` as the container<=>host propagation path will be the same

MrCry0: On the other side, `kpartx` is a modern tool and may use a more reliable way of creating/checking loop devices than losetup.
We could check it.

doanac: kpartx looks interesting. I tried to google "kpartx vs losetup" to find some differences and don't get much. I say for now we merge this as a small improvement. and we can revisit kpartx in the future if we continue to have sporadic issues.

assemble.py
doanac: 943a76dde4640b7750955ad0cbec68529d911333

PR #216, opened 2022-10-05
apps: Build containers with two labels

doanac:
We have a problem with factories using a common base image as described:

 https://docs.foundries.io/88/reference-manual/docker/containers.html#id1

Currently, every build is causing the images that extend the base image to be "rebuilt". I say that in quotes because its still the same image: every single layer the image is made from remains the same. However, the docker manifest pointing to the layers does change.

This comes from my misunderstanding of the "FROM" line in the child containers. Our docs have have basically inherit from base:${tag}-${arch}. ${tag} is changing every single build. While the SHA256 it points to does not change, the logic of a Dockerfile to decide what to build more or less does a string equals comparison of each line of the Dockerfile. This causes things to rebuild.

We instead need the child containers to have a consistent FROM line:

 ${LATEST}-${arch}

This tag isn't changing but its *sha* will when the base image changes.

This fix moves the logic to evaluate LATEST before we source pre-build.conf and then have ci publish two tags for each container we build:

 ${tag}-${arch}    - what we've always had
 ${LATEST}-${arch} - the one we need for base images

NOTE: This allows things to work we still have to fix:

 * factories that use this. Pre-build.conf needs a change to have: `_base_img="hub.foundries.io/${FACTORY}/0base:$LATEST-$ARCH"`

 * documentation that fixes the pre-build.conf template

### DIFF
--- a/apps/build.sh
+++ b/apps/build.sh
@@ -51,6 +51,10 @@ if [ -n "$DOCKER_BUILDX" ] ; then
 	docker_build="docker buildx build"
 fi
 
+TAG=$(git log -1 --format=%h)
+LATEST=${OTA_LITE_TAG-"latest"}
+LATEST=$(echo $LATEST | cut -d: -f1 | cut -d, -f1)  # Take into account advanced tagging
+
 pbc=pre-build.conf
 if [ -f $pbc ] ; then
   echo "Sourcing pre-build.conf."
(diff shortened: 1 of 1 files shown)


doanac:
github renders my commit message in an unusual way.

vanmaegima (approved):

vanmaegima:
@doanac This works fine and the container sha doesn't change anymore. I noticed that it is still being re-published, but I don't know if this is an issue or not: https://hub.foundries.io/factories/stm32-vanessa/system-monitor (builds 59 -> 61).

The only change in my current pre-build.conf file was:
_base_img="hub.foundries.io/${FACTORY}/0base-image:$LATEST-$ARCH"

Build log from a dummy change in the second app for reference: https://app.foundries.io/factories/stm32-vanessa/targets/61/runs/build-armhf/console

The container ID for the unchanged app stays the same.

I also tested a base image change and all apps rebuild.

Anything else I should test?

doanac:
>  I noticed that it is still being re-published,

We do that or then you get into a weird support issue where people ask "where is container 61?" and then we have to go back through CI jobs and git history to figure out that 61 is actually 59 in your case.

doanac:
I tested it pretty well too with:
 * change a random container (not the base or child) - nothing changed in those
 * change the base - everything changed
 * changed one child - the other child didn't change
 
I think we can merge this.

PR #218, opened 2022-10-13
lmp/bb-config: expose IMAGE as LMP_FACTORY_IMAGE

ricardosalveti:
IMAGE is handled in a special way in OE/Yocto, so it can't be easily consumed by recipes.

In order for the image name to be available for other recipes to consume (e.g. os-release), create an additional variable that is set to the image name given by the factory ci job.


quaresmajose (approved):
LGTM

mike-scott (approved):
LGTM

PR #221, opened 2022-10-20
Build and deploy the Yocto Project Standard SDK (toolchain)

angolini:


lmp/bb-build.sh
```diff
@@ -18,4 +18,8 @@ bitbake -e ${IMAGE} > ${archive}/bitbake_image_env.txt
 # Setscene (cache), failures not critical
 bitbake --setscene-only ${IMAGE} || true
 
-bitbake -D ${BITBAKE_EXTRA_ARGS} ${IMAGE}
+if [ "$BUILD_SDK" == "1" ] && [ "${DISTRO}" != "lmp-mfgtool" ]; then
+    bitbake -D ${BITBAKE_EXTRA_ARGS} ${IMAGE} -c populate_sdk
+else
+    bitbake -D ${BITBAKE_EXTRA_ARGS} ${IMAGE}
+fi
```
quaresmajose, line 25: This will require a particular build for the sdk target, that can be inconvenient in some uses cases.
For example, if the user needs to generate an image with the corespondent sdk they will need to run 2 builds, one for the image and another for the sdk.
IMO we can run always the build image and when needed the sdk, that only run the task populate_sdk with all the sstate cache collected from the image build process.
```
bitbake -D ${BITBAKE_EXTRA_ARGS} ${IMAGE}
if [ "$BUILD_SDK" == "1" ] && [ "${DISTRO}" != "lmp-mfgtool" ]; then
    bitbake -D ${BITBAKE_EXTRA_ARGS} ${IMAGE} -c populate_sdk
fi
```

angolini: Yes, you are right @quaresmajose 

but as I understand, this point is still being decided. I will leave this comment open until we make the final decision.

ricardosalveti: Yeah, I think we can do as suggested by @quaresmajose for now, until we decide the final API for it. That will at least allow customers to start generating the sdk automatically when required.

angolini: @quaresmajose @ricardosalveti done

quaresmajose (approved):
LGTM

PR #222, opened 2022-10-21
apps: Handle git-sha collision bug in tagging

doanac:
Our logic was hard-coded to assume Git SHA's are always 7 long. Git SHAs can sometimes collide in the first 7 characters leading the hash to be longer.

This change dynamically determines the SHA length so that we can handle all input.


Tim-Anderson (approved):
It does not look like it will break anything. LGTM

PR #227, opened 2022-10-25
helpers: repo: temporarily use main for another attempt to fix 16355

ricardosalveti:
Fixed v2.29.2 had issues when doing repo sync inside the container.


Tim-Anderson (approved):
That seems to do the trick

PR #231, opened 2022-11-03
Enable buildx explicitly and remove legacy logic

doanac:
Our logic for building containers predates buildx. We eventually started doing buildx and moving customers over one by. Turns out, we didn't migrate one of our oldest customers. They are now needing something buildx offers. This change set makes us use buildx. It's safe because everyone else was already on it. While enforcing buildx, I was able to remove a lot of old conditional logic that simply can't be run now.

### DIFF
--- a/apps/build.sh
+++ b/apps/build.sh
@@ -117,49 +117,44 @@ for x in $IMAGES ; do
 		auth=1
 	fi
 
-	if [ $no_op_tag -eq 1 ] ; then
-		status Tagging docker image $x for $ARCH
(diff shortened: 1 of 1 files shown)


doanac:
@mike-sul - i can imagine this could be a pain to read since you weren't around with @EmbeddedAndroid and I in the old days. I've tried to make small changes that are fairly obvious to verify. Good luck

mike-sul (approved):
LGTM. I suppose it makes sense to test it e2e in factory by running CI builds.

doanac:
Its in my factory and the affected customers factory already. They are just wanting it merged so they can get back on our main branch.

PR #235, opened 2022-11-17
LmP: refactor INHERIT build config

quaresmajose:
**Buildstats**
The buildstats option is only enabled on the CI build but it is very useful in local builds as well to check sstate-cache usage.

**Create-spdx**
The create-spdx is been moved to only runs on CI as local builds can run in the majority of time without this
that reduces the build time.

### DIFF
--- a/lmp/bb-config.sh
+++ b/lmp/bb-config.sh
@@ -229,8 +229,6 @@ EOFEOF
 fi
 
 cat << EOFEOF >> conf/auto.conf
-# get build stats to make sure that we use sstate properly
-INHERIT += "buildstats buildstats-summary"
 
 # archive sources for target recipes (for license compliance)
 INHERIT += "archiver"
--- a/lmp/bb-config.sh
+++ b/lmp/bb-config.sh
@@ -237,6 +237,15 @@ ARCHIVER_MODE[src] = "original"
 ARCHIVER_MODE[diff] = "1"
 EOFEOF
 
+# spdx is support since kirkstone so check if this oe-core have support
+if [ -f ../layers/openembedded-core/meta/classes/create-spdx.bbclass ]; then
+	cat << EOFEOF >> conf/auto.conf
+
+# create SPDX (SBOM) documents
+INHERIT += "create-spdx"
+EOFEOF
+fi
+
 if [ $(ls ../sstate-cache | wc -l) -ne 0 ] ; then
 	status "Found existing sstate cache, using local copy"
 	echo 'SSTATE_MIRRORS = ""' >> conf/auto.conf


lmp/bb-config.sh
```diff
@@ -218,14 +218,14 @@ EOFEOF
...
-INHERIT += "buildstats buildstats-summary"
-
 # archive sources for target recipes (for license compliance)
 INHERIT += "archiver"
 COPYLEFT_RECIPE_TYPES = "target"
 ARCHIVER_MODE[src] = "original"
 ARCHIVER_MODE[diff] = "1"
+
+# create SPDX (SBOM) documents
+INHERIT += "create-spdx"
```
ricardosalveti, line 228: The problem here, with the removal and addition, is that it will affect every factory we have, and they are all based on different LmP revisions (some which might not have this create-spdx class). Do you know if bitbake will complain when this class is not available?

quaresmajose: I did a quick test here on kiskstone with  ` INHERIT += "create-spdx1"` and bitbake didn't complain. Maybe on old version we get the same but need to check, at last on dunfell.

lmp/bb-config.sh
```diff
@@ -218,14 +218,14 @@ EOFEOF
 fi
 
 cat << EOFEOF >> conf/auto.conf
-# get build stats to make sure that we use sstate properly
-INHERIT += "buildstats buildstats-summary"
```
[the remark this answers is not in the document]
quaresmajose: yeah, need to check as well. if we are fine with `create-spdx` we are fine here too.

angolini:
Is this the first time the local build and CI build will differ?

I agree and understand the PR goal. But let us be cautious. At least let's have an internal page with all the differences between local build and CI or any other doc.

@kprosise, could you please, add this PR to your radar?

lmp/bb-config.sh
quaresmajose: Unfortunately doesn't work on older versions
```
builder@c9847f2d25db:/lmp/build-intel-corei7-64-lmp$ bitbake lmp-base-console-image -p
ERROR: Unable to start bitbake server (None)
ERROR: Server log for this session (/lmp/build-intel-corei7-64-lmp/bitbake-cookerdaemon.log):
--- Starting bitbake server pid 274 at 2022-11-17 16:44:55.702530 ---
ERROR: ParseError in configuration INHERITs: Could not inherit file classes/create-spdx.bbclass
```

So it need to be conditioned with something like
```
if [ -f layers/openembedded-core/meta/classes/sstate.bbclass ] ; then
    echo 'INHERIT += "create-spdx"' >> conf/auto.conf
fi
```

lmp/bb-config.sh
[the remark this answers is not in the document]
quaresmajose: this one is less problematic and will don't cause any error and at the end if they will be necessary it can be added in some customer layers.

lmp/bb-config.sh
quaresmajose: > I did a quick test here on kiskstone with ` INHERIT += "create-spdx1"` and bitbake didn't complain. Maybe on old version we get the same but need to check, at last on dunfell.

Don't know why but this is not true! checking again on kiskstone and it fails as in dunfell
```
ERROR: ParseError in configuration INHERITs: Could not inherit file classes/create-spdx1.bbclass
```

quaresmajose: There are some movements upstream to backport the `create-spdx` to the LTS dunfell branhc that will be a reality very soon.

ricardosalveti: The problem for us is that we need support in OE and then make sure that every factory is using an updated lmp release.

quaresmajose: I understand and that is the reason to enable the spdx only when the create-spdx.bbclass exist on the layer.

ricardosalveti: Yup, sounds good.

ricardosalveti (approved):
LGTM, we should coordinate the landings to make sure we don't introduce any regression.

quaresmajose:
We can merge this one and if the `buildstats` is still required in older version of any costumer we can enable only that in the `meta-subscriber-overrides`.

PR #239, opened 2022-11-30
Sstate cache updates

doanac:


doanac:
hey guys -no rush on this review. i want to deploy after v89. However, it has some interesting changes for both the public lmp factory and regular users.

For the public lmp factory we: ill use new workers in GCP that:
 * don't use NFS
 * will use an HTTPS sstate cache mirror backed by a GCS bucket
 * sync their cache (done inside jobserv worker logic) after LmP builds
 * produce a versioned cache (we use current LmP tag +1 since we are always working on the next release until I tag things)

For customers we'll:
 * GCSFuse mounted cache of what LmP builds above
 * produce versioned caches so its easier for us to prune old stuff on their NFS server(s) 

I've been doing old v88 release builds and andy-corp builds on this, but its hard to truly test it until we move public lmp over. The customer logic should be fine either way - it handles migration logic to do the right thing so that the timing of the rollout doesn't matter.

lmp/jobserv.yml
```diff
@@ -615,6 +615,7 @@ params:
   DISTRO: lmp
   SOTA_CLIENT: aktualizr-lite
   OSTREE_BRANCHNAME: lmp
+  SSTATE_CACHE_MIRROR: https://storage.googleapis.com/lmp-cache
```
quaresmajose, line 618: Serve the public LmP by version is something that will not scale well, this will not work transparently for the end user without change this file https://github.com/foundriesio/lmp-manifest/pull/255/files for each LmP release.
My suggestion is merging all the sstate folders with overlayfs and made this available on the http server.

doanac: i don't control the implementation of `https://storage.googleapis.com/lmp-cache` so i can't do an overlay for that. And we've learned that we can't control the size of sstate cache w/o versioning it. It will simply grow unbounded over time.

quaresmajose: I understand and agree with versioning the cache  but I am thinking in a clean way of sharing the cache.  Looking again on that and I can add some logic on the https://github.com/foundriesio/lmp-manifest/pull/255 to handle the version without user intervention so we can avoid the overlay for now.

ricardosalveti: A similar approach can be done in setup-environment when generating auto.conf or similar, to use the correct path.

ricardosalveti (commented):
LGTM, not that intrusive, so merge as you get to play with this better.

ricardosalveti (approved):

lmp/jobserv.yml
quaresmajose: yeah have already add it

quaresmajose (approved):
LGTM

lmp/jobserv.yml
doanac: I think the problem is that our CI scripts will default this to what we want for customers - https://github.com/foundriesio/ci-scripts/blob/master/lmp/bb-config.sh#L20 - so we have to *overrride* that for CI.

quaresmajose: We can use one or the other in the CI so we can rewrite the https://github.com/foundriesio/ci-scripts/pull/239/files#diff-f12b382648c4eb1512e3828fbc501e21e138c7b08fba2f6d74410c22ae84bf4aR207-R216

```
if [[ "$SSTATE_CACHE_MIRROR" == "https://"* ]]  ; then
	cat << EOFEOF >> conf/local.conf
SSTATE_MIRRORS = "file://.* ${SSTATE_CACHE_MIRROR}/v$LMP_VER-sstate-cache/PATH"
EOFEOF
elif [ -d $SSTATE_CACHE_MIRROR ]; then
	cat << EOFEOF >> conf/local.conf
SSTATE_MIRRORS = "file://.* file://${SSTATE_CACHE_MIRROR}/PATH"
EOFEOF
fi
```
and this will prioritize the http mirror over the nfs

lmp/bb-config.sh
```diff
@@ -205,6 +209,11 @@ if [ -d $SSTATE_CACHE_MIRROR ]; then
 SSTATE_MIRRORS = "file://.* file://${SSTATE_CACHE_MIRROR}/PATH"
 EOFEOF
 fi
+if [[ "$SSTATE_CACHE_MIRROR" == "https://"* ]]  ; then
+	cat << EOFEOF >> conf/local.conf
+SSTATE_MIRRORS = "file://.* ${SSTATE_CACHE_MIRROR}/v$LMP_VER-sstate-cache/PATH"
+EOFEOF
+fi
```
quaresmajose, line 216: We can prioritize the http mirror over the nfs
```
if [[ "$SSTATE_CACHE_MIRROR" == "https://"* ]]  ; then
	cat << EOFEOF >> conf/local.conf
SSTATE_MIRRORS = "file://.* ${SSTATE_CACHE_MIRROR}/v$LMP_VER-sstate-cache/PATH"
EOFEOF
elif [ -d $SSTATE_CACHE_MIRROR ]; then
	cat << EOFEOF >> conf/local.conf
SSTATE_MIRRORS = "file://.* file://${SSTATE_CACHE_MIRROR}/PATH"
EOFEOF
fi
```

lmp/jobserv.yml
doanac: yes, but the problem will still remain that unless I set it to https://.... here, it will get defaulted to a directory above with:
```
SSTATE_CACHE_MIRROR="${SSTATE_CACHE_MIRROR-/sstate-cache-mirror/v$LMP_VER-sstate-cache}"
```
and in the case of CI - its probably going to be a one or the other.

quaresmajose: another approach, is taking the value defined by the setup-environment-internal https://github.com/foundriesio/lmp-manifest/pull/255, it will adds the http mirros as a weak default so it will be used for all build if no other configuration exist.
In this approach we need to setup the nfs mirror (to the directory) for the non-public lmp builds on the ci.

doanac: I think I'm missing something or thinking about this wrong. Maybe you can help show me in this PR how this can be simplified. The problem I have is that every single customer uses this script and doesn't not have `SSTATE_CACHE_MIRROR` defined. So I *think* I need to default this to what they all currently need `/sstate-cache-mirror/v$LMP_VER-sstate-cache`. However, for this public LmP CI job we need this to not be via HTTP. 

Also - as I commented in your PR - the the value of LMP_VER for this and that should be two different things?

doanac:
rebased on latest master

PR #240, opened 2022-12-16
lmp: Get version tag from LmP repository

doanac:
Some customers may put tags in their lmp-manifest repo which will can cause us to get the wrong cache value


quaresmajose (approved):
LGTM

quaresmajose:
Do you think we need this on the public lmp https://github.com/foundriesio/lmp-manifest/pull/255 so it can work with forked repos?

doanac:
probably so.

PR #241, opened 2022-12-17
lmp: Fix logic for detecting the base version

doanac:
Its a little tricky to pick up the base LmP version. We need to make sure we've fetched tags from the public LmP and *don't* look at tags that have been created by the user.

The old version was also getting the *latest* tag which is currently v89 but that's probably not the actual LmP version the customer is building with. Git doesn't have a way to distinguish where tags come from, so the easiest thing to do is just delete all the tags and then pull from our official tags.

### DIFF
--- a/helpers.sh
+++ b/helpers.sh
@@ -73,3 +73,31 @@ function repo_sync {
 		cp .repo/manifest.xml $archive/manifest.xml
 	fi
 }
+
+function set_base_lmp_version {
+	# Detect the base LmP version we are building on
+	# A user may have place arbitrary tags into their lmp-manifest and we
+	# don't want to accidentally pick one of them up. Git doesn't really
+	# distinguish where tags come from, so:
+
+	# 1: Delete all tags
+	git --git-dir .repo/manifests.git tag | xargs git --git-dir .repo/manifests.git tag -d > /dev/null
+
+	# 2: Get the LMP tags
+	run git --git-dir .repo/manifests.git remote add upstream https://github.com/foundriesio/lmp-manifest
+	run git --git-dir .repo/manifests.git fetch --tags upstream
+
+	# 3: Find our base LMP version based on the HEAD
+	export LMP_VER=$(git --git-dir .repo/manifests.git describe HEAD --tags --abbrev=0)
+	if [[ "${H_PROJECT}" == "lmp" ]] ; then
+		# Public LmP build - we are building for the *next* release
+		LMP_VER=$(( $LMP_VER + 1 ))
+	fi
+
+	# 4: Delete all tags again
+	git --git-dir .repo/manifests.git tag | xargs git --git-dir .repo/manifests.git tag -d > /dev/null
+
+	# 5: Restore orginal tags - the build may need them
+	run git --git-dir .repo/manifests.git fetch --tags
+	status "Base LmP version detected as: $LMP_VER"
+}
--- a/lmp/build.sh
+++ b/lmp/build.sh
@@ -32,15 +32,8 @@ fi
 cat /root/.gitconfig >>  /home/builder/.gitconfig
 cp /root/.netrc /home/builder/.netrc || true
 
-# Detect the base LmP version we are building on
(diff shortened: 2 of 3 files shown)


doanac:
I discovered this today after looking at some customer CI runs. Everything works for now, but once they try and start pulling a versioned public LmP cache (we currently are using fallback logic to the old cache) - they could hit the wrong cache version and miss most cache hits.

NOTE - I've been keeping a list of factories that have done LmP builds. i'll need to do a small manual migration after this is merged to correct their cache versions.

lmp/build.sh
```diff
@@ -33,9 +33,13 @@ cat /root/.gitconfig >>  /home/builder/.gitconfig
...
 
 # Detect the base LmP version we are building on
+# A user may have place arbitrary tags into their lmp-manifest and we don't
+# want to accidentally pick one of them up. Git doesn't really distinguish
+# were tags come from, so:
+git --git-dir .repo/manifests.git tag | xargs git --git-dir .repo/manifests.git tag -d > /dev/null
 run git --git-dir .repo/manifests.git remote add upstream https://github.com/foundriesio/lmp-manifest
 run git --git-dir .repo/manifests.git fetch --tags upstream
-export LMP_VER=$(git --git-dir .repo/manifests.git describe upstream/main --tags --abbrev=0)
+export LMP_VER=$(git --git-dir .repo/manifests.git describe HEAD --tags --abbrev=0)
```
quaresmajose, line 42: We can restore the costumer repo at the end
```
run git --git-dir .repo/manifests.git fetch --tags origin
run git --git-dir .repo/manifests.git  remote upstream
```

lmp/build.sh
```diff
@@ -33,9 +33,13 @@ cat /root/.gitconfig >>  /home/builder/.gitconfig
 cp /root/.netrc /home/builder/.netrc || true
 
 # Detect the base LmP version we are building on
+# A user may have place arbitrary tags into their lmp-manifest and we don't
+# want to accidentally pick one of them up. Git doesn't really distinguish
+# were tags come from, so:
+git --git-dir .repo/manifests.git tag | xargs git --git-dir .repo/manifests.git tag -d > /dev/null
```
quaresmajose, line 39: Why this git operations don't use the helper `run` like the others?

doanac: it was just subjective. "run" shows the command and ouput but i'm not sure it works great with a piped sequence. i think there's enough logic stacking up here, that I'm going to make a shell function to help contain it.

lmp/build.sh
doanac: good catch. I did a force push that moves things to a helper function and I wrote a unit test.

quaresmajose (approved):
LGTM

helpers.sh
```diff
@@ -73,3 +73,28 @@ function repo_sync {
...
+
+	# 3: Find our base LMP version based on the HEAD
+	export LMP_VER=$(git --git-dir .repo/manifests.git describe HEAD --tags --abbrev=0)
+	if [[ "${H_PROJECT}" == "lmp" ]] ; then
+		# Public LmP build - we are building for the *next* release
+		LMP_VER=$(( $LMP_VER + 1 ))
+	fi
+
+	# 4: Restore orginal tags - the build may need them
+	run git --git-dir .repo/manifests.git fetch --tags
```
ricardosalveti, line 98: At this stage do we need to drop the upstream remote and the tags associated with it? Thinking on possible tag conflicts and what this command would do in case we have conflicts.

I don't think it is something we should be worried with, but thinking on unexpected side effects.

doanac: can't hurt. I've force pushed this idea and am testing in my factory

doanac: its working. i think i can merge now if you are fine with this. then i can manually fix some old factories to point at the correct version

ricardosalveti: lgtm

ricardosalveti (approved):
LGTM

PR #242, opened 2023-01-06
helpers: set_base_lmp_version: work on a copy of the git repo

quaresmajose:

### DIFF
--- a/helpers.sh
+++ b/helpers.sh
@@ -86,7 +86,8 @@ function set_base_lmp_version {
 	git fetch origin --tags --quiet
 
 	# 3: Find our base LMP version based on the HEAD
-	export LMP_VERSION_CACHE=$(git describe --tags --abbrev=0 HEAD)
+	export LMP_VERSION=$(git describe --tags --abbrev=0 HEAD)
+	export LMP_VERSION_CACHE="$LMP_VERSION"
 	if [[ "${H_PROJECT}" == "lmp" ]] ; then
 		# Public LmP build - we are building for the *next* release
 		LMP_VERSION_CACHE=$(( $LMP_VERSION_CACHE + 1 ))
--- a/helpers.sh
+++ b/helpers.sh
@@ -86,12 +86,12 @@ function set_base_lmp_version {
 	git fetch origin --tags --quiet
 
 	# 3: Find our base LMP version based on the HEAD
-	export LMP_VER=$(git describe --tags --abbrev=0 HEAD)
+	export LMP_VERSION_CACHE=$(git describe --tags --abbrev=0 HEAD)
 	if [[ "${H_PROJECT}" == "lmp" ]] ; then
 		# Public LmP build - we are building for the *next* release
-		LMP_VER=$(( $LMP_VER + 1 ))
+		LMP_VERSION_CACHE=$(( $LMP_VERSION_CACHE + 1 ))
 	fi
-	status "Base LmP version detected as: $LMP_VER"
+	status "Base LmP cache version detected as: $LMP_VERSION_CACHE"
 
 	# 4: cleanup
 	popd >/dev/null
--- a/helpers.sh
+++ b/helpers.sh
@@ -75,29 +75,25 @@ function repo_sync {
 }
 
 function set_base_lmp_version {
-	# Detect the base LmP version we are building on
-	# A user may have place arbitrary tags into their lmp-manifest and we
-	# don't want to accidentally pick one of them up. Git doesn't really
-	# distinguish where tags come from, so:
+	# 1: Work on a copy
+	WORKCOPY="$PWD/.repo/manifests.git.tmp"
+	cp -R .repo/manifests.git $WORKCOPY
+	pushd $WORKCOPY >/dev/null
 
-	# 1: Delete all tags
-	git --git-dir .repo/manifests.git tag | xargs git --git-dir .repo/manifests.git tag -d > /dev/null
-
-	# 2: Get the LMP tags
-	run git --git-dir .repo/manifests.git remote add upstream https://github.com/foundriesio/lmp-manifest
-	run git --git-dir .repo/manifests.git fetch --tags upstream
+	# 2: Replace all tags
+	git tag --delete $(git tag) >/dev/null
+	git remote set-url origin https://github.com/foundriesio/lmp-manifest
+	git fetch origin --tags --quiet
 
 	# 3: Find our base LMP version based on the HEAD
-	export LMP_VER=$(git --git-dir .repo/manifests.git describe HEAD --tags --abbrev=0)
+	export LMP_VER=$(git describe --tags --abbrev=0 HEAD)
 	if [[ "${H_PROJECT}" == "lmp" ]] ; then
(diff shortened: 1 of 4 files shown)


doanac:
I'm re-running the build https://app.foundries.io/factories/lmp/targets/1979/runs/build-intel-corei7-64/console with your change.

doanac:
as noted in slack. This fix is fine, but the real issue is that https://github.com/foundriesio/lmp-manifest/commit/d7a019b2297ed92f31399220facccb10139641e8 is overriding what we do here.

quaresmajose:
> as noted in slack. This fix is fine, but the real issue is that [foundriesio/lmp-manifest@d7a019b](https://github.com/foundriesio/lmp-manifest/commit/d7a019b2297ed92f31399220facccb10139641e8) is overriding what we do here.

 I already knew that `setup-environment` was called in `lmp/bb-build.sh:9` but I didn't know that it was also in `lmp/bb-config.sh:51` and this one will massed up the CI config written in `lmp/bb-config.sh`
 
```
grep setup-environment -Rn
lmp/bb-build.sh:9:source setup-environment build
lmp/bb-config.sh:51:source setup-environment build
```

helpers.sh
```diff
@@ -86,12 +86,12 @@ function set_base_lmp_version {
 	run git fetch origin --tags --quiet
 
 	# 3: Find our base LMP version based on the HEAD
-	export LMP_VER=$(git describe --tags --abbrev=0 HEAD)
+	export LMP_CACHE_VERSION=$(git describe --tags --abbrev=0 HEAD)
```
quaresmajose, line 90:
```
-LMP_VER
+LMP_VERSION_SHORT
```
maybe this one is better

lmp/bb-config.sh
```diff
@@ -8,7 +8,7 @@ OSTREE_BRANCHNAME="${OSTREE_BRANCHNAME-lmp-localdev}"
 SOTA_CLIENT="${SOTA_CLIENT-aktualizr}"
 AKLITE_TAG="${AKLITE_TAG-promoted}"
 H_BUILD="${H_BUILD-lmp-localdev}"
-LMP_VERSION=$(git --git-dir=.repo/manifests/.git describe --tags 2>/dev/null || echo unknown)
+LMP_VERSION="$(LMP_VERSION-unknown)"
```
MrCry0, line 11: The same here: `"$(LMP_VERSION:-unknown)"` is better to prevent passing forward empty value.

helpers.sh
```diff
@@ -75,29 +75,26 @@ function repo_sync {
...
-	run git --git-dir .repo/manifests.git remote add upstream https://github.com/foundriesio/lmp-manifest
-	run git --git-dir .repo/manifests.git fetch --tags upstream
+	# 2: Replace all tags
+	run git tag --delete $(git tag) >/dev/null
+	run git remote set-url origin https://github.com/foundriesio/lmp-manifest
+	run git fetch origin --tags --quiet
 
 	# 3: Find our base LMP version based on the HEAD
-	export LMP_VER=$(git --git-dir .repo/manifests.git describe HEAD --tags --abbrev=0)
+	export LMP_VERSION=$(git describe --tags HEAD)
```
ricardosalveti, line 89: I wonder if it would just be better to also fix LMP_VERSION to use abbrev=0 as well, making it the same as LMP_CACHE_VERSION.

Since what we really want here is the tag revision.

helpers.sh
doanac: yep. I hate to nitpick variable names, but I think `LMP_VERSION` (since it was already that) and `LMP_VERSION_SHORT` or `LMP_VERSION_NUM` would be better.

lmp/bb-config.sh
quaresmajose: I understand you suggestion but using here `:-` when all of the others only uses `-`  doesn't look very nice

helpers.sh
quaresmajose: we need to use two variables for this because the version used on the sstate-cache can be incremented.

helpers.sh
quaresmajose: changed to `LMP_VERSION_NUM`

doanac (approved):
Looks good to me

MrCry0 (approved):
LGTM

helpers.sh
ricardosalveti: While now it is easy to understand, I find reusing these variables a bit confusing and I'm sure it will cause confusion later on once we have to change this codebase again.

Why don't we just using one single LMP_VERSION variable (fetching with abbrev=0) and just define LMP_VERSION_CACHE instead of VERSION_NUM (which can be confusing).

That way we set LMP_VERSION as the global variable to be used to find out the version the build is based on, and LMP_VERSION_CACHE used to point to the cache version/bucket to be used by the build instead.

quaresmajose: Change LMP_VERSION variable (fetching with abbrev=0) can impact the target roofs as it will change `os-release`, is this acceptable? 

We can maintain the LMP_VERSION as before (i.e `89-27-g8f931b6`) and strip down only the release number with cut, storing the result on  LMP_VERSION_CACHE that can be incremented and will be used for sstate cache.
```
# LMP_VERSION=89-27-g8f931b6
LMP_VERSION_CACHE="$(echo $LMP_VERSION | cut -d'-' -f1)"
```

ricardosalveti: That is fine as every CI job will have a customized version field as it contains the build number as well.

ricardosalveti: And my suggestion is because we should have used abbrev=0 when we did this first implementation.

quaresmajose: Fixed

ricardosalveti (approved):
Nice, thanks! We just need to validate this on a factory and we should be good to land.

quaresmajose:
> Nice, thanks! We just need to validate this on a factory and we should be good to land.

I will run a build with this in one of my factories.

quaresmajose:
> > Nice, thanks! We just need to validate this on a factory and we should be good to land.
> 
> I will run a build with this in one of my factories.

Remove the last tag `89` first from my repo and it is still detect as expected from the upstream version

```
/srv/oe/.repo/manifests.git.tmp /srv/oe
== 2023-01-10 18:40:10 Running: git remote set-url origin https://github.com/foundriesio/lmp-manifest
|--
== 2023-01-10 18:40:10 Running: git fetch origin --tags --quiet
|--
== 2023-01-10 18:40:11 Base LmP cache version detected as: 89
/srv/oe
```
I will cleanup it a bit removing  all these debug commands and let's just stick with `== 2023-01-10 18:40:11 Base LmP cache version detected as: 89
`

Anyawy we can't test the http mirror on the factory because they are used only in public lmp builds, on the factory we use the previous mechanism over nfs
```
SSTATE_MIRRORS = "file://.* file:///sstate-cache-mirror/PATH"
```

ricardosalveti:
> Anyawy we can't test the http mirror on the factory because they are used only in public lmp builds, on the factory we use the previous mechanism over nfs

That's fine, not so concerned with the public lmp jobs.

PR #243, opened 2023-01-10
apps: Set default targets expiration to one month

mike-sul:
The newer version of `garage-sign` does not refresh targets meta expiration by default as it's used to be in the older versions. So, let's set the expiration value to one month explicitly in the `garage-sign` command call.
The same already happens in the LmP build too, see `meta-updater/classes/image_types_ostree.bbclass`.


doanac (approved):
thanks

PR #244, opened 2023-01-11
Adds support to use the HTTP sstate cache mirror on the factories

quaresmajose:
The variable `SSTATE_CACHE_MIRROR` is to force the use of the public LmP HTTP ssate cache mirror: 
```
# factory-config.yml

  params:
    IMAGE: lmp-factory-image
    SSTATE_CACHE_MIRROR: ""
```

The variable `LMP_VERSION_CACHE_DEV` is to use the development version of the cache: 
```
# factory-config.yml

  params:
    IMAGE: lmp-factory-image
    SSTATE_CACHE_MIRROR: ""
    LMP_VERSION_CACHE_DEV: ""
```

### DIFF
--- a/helpers.sh
+++ b/helpers.sh
@@ -94,7 +94,7 @@ function set_base_lmp_version {
 	# 3: Find our base LMP version based on the HEAD
 	export LMP_VERSION=$(git describe --tags --abbrev=0 HEAD)
 	export LMP_VERSION_CACHE="$LMP_VERSION"
-	if [[ "${H_PROJECT}" == "lmp" ]] ; then
+	if [[ "${H_PROJECT}" == "lmp" ]] || [ -v LMP_VERSION_CACHE_DEV ] ; then
 		# Public LmP build - we are building for the *next* release
 		LMP_VERSION_CACHE=$(( $LMP_VERSION_CACHE + 1 ))
 	fi
--- a/lmp/bb-config.sh
+++ b/lmp/bb-config.sh
@@ -211,11 +211,6 @@ if [ -d "$SSTATE_CACHE_MIRROR" ]; then
 SSTATE_MIRRORS = "file://.* file://${SSTATE_CACHE_MIRROR}/PATH"
 EOFEOF
 fi
-if [[ "$SSTATE_CACHE_MIRROR" == "https://"* ]]  ; then
-	cat << EOFEOF >> conf/local.conf
-SSTATE_MIRRORS = "file://.* ${SSTATE_CACHE_MIRROR}/v$LMP_VERSION_CACHE-sstate-cache/PATH"
-EOFEOF
-fi
 
 # Add build id H_BUILD to output files names
 if [ "$CONF_VERSION" == "1" ]; then
--- a/lmp/bb-config.sh
+++ b/lmp/bb-config.sh
@@ -208,7 +208,7 @@ fi
 
 if [ -d "$SSTATE_CACHE_MIRROR" ]; then
 	cat << EOFEOF >> conf/local.conf
-SSTATE_MIRRORS = "file://.* file://${SSTATE_CACHE_MIRROR}/PATH"
+SSTATE_MIRRORS ?= "file://.* file://${SSTATE_CACHE_MIRROR}/PATH"
 EOFEOF
 fi
 
--- a/lmp/bb-config.sh
+++ b/lmp/bb-config.sh
@@ -241,11 +241,6 @@ INHERIT += "create-spdx"
 EOFEOF
 fi
 
-if [ $(ls ../sstate-cache | wc -l) -ne 0 ] ; then
-	status "Found existing sstate cache, using local copy"
-	echo 'SSTATE_MIRRORS = ""' >> conf/auto.conf
-fi
-
 for x in $(ls conf/*.conf) ; do
 	status "$x"
 	cat $x | indent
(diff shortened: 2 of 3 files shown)


lmp/bb-config.sh
```diff
@@ -236,11 +240,6 @@ ARCHIVER_MODE[src] = "original"
 ARCHIVER_MODE[diff] = "1"
 EOFEOF
 
-if [ $(ls ../sstate-cache | wc -l) -ne 0 ] ; then
-	status "Found existing sstate cache, using local copy"
-	echo 'SSTATE_MIRRORS = ""' >> conf/auto.conf
-fi
```
doanac, line 243: Can you elaborate what this change is for in the commit message, I don't quite understand this. 

I'd think this specific line of code would be needed or our customer factory builds would break?

quaresmajose: I am assuming this is the local cache of the factory. That just doesn't exist in the first build, being populated after this so this `../sstate-cache` folder always have some content. Then the sstate cache will always be off after the first build.

quaresmajose: I need to look on the job server code to see what is mounted on this `../sstate-cache` folder

quaresmajose: I have checked this snip `Found existing sstate cache, using local copy` and it is used always after the first build.
So it is what I am think about, our factories only use the mirror for the first build and after `SSTATE_MIRRORS = ""`.
I will update to reflect that.

quaresmajose: Actually I think this is not correct and we should always use SSTATE_MIRRORS so the factories can take advantage of the public cache when updating from v(N-1) to v(N).

Bitbake will check for sstate cache in the following order:

1. SSTATE_DIR
2. SSTATE_MIRRORS

So it will only search the mirror if the file not found on the local cache `../sstate-cache`, this is an insignificant overhead searching the `SSTATE_MIRRORS` because most of the files are on the local `SSTATE_DIR`

ricardosalveti: So should we change to always use the http mirror by default? Currently an elif.

doanac: that's about right. When using NFS (aka "persistent jobserv volumes"), we don't want to use HTTP because it causes duplicate copies of all the base LmP objects. Makes each factory about 7G more multiplied by `num_customers`.

for our "nocache" builds (like the LmP) they don't use persistent volumes, so HTTP works pretty well. *and* our public LmP builds are done on CI workers that copy the sstate cache generated locally back to the storage bucket (where HTTP is served from).

quaresmajose: Right, so with this PR we add the options to change the default mirror used in the factory.

By default the mirror used is the local one that is the shared NFS mounted on the container:
https://github.com/foundriesio/ci-scripts/blob/f56cc8b494a5882a30471100d44b95bcd605f038/lmp/bb-config.sh#L20

The http mirror is disabled in costumer factories and is enabled only on public lmp with:
https://github.com/foundriesio/ci-scripts/blob/f56cc8b494a5882a30471100d44b95bcd605f038/lmp/jobserv.yml#L626

With the above same change in any costumer factory the user can change the mirror

quaresmajose: the only missing part is the increment of the `LMP_VERSION_CACHE` for the development release. Maybe we need the same `LMP_VERSION_CACHE_DEV` on ci too.
https://github.com/foundriesio/ci-scripts/blob/f56cc8b494a5882a30471100d44b95bcd605f038/helpers.sh#L97-L100

quaresmajose: Also add the `LMP_VERSION_CACHE_DEV` so everything is in place to use the HTTP mirror on the factories

quaresmajose:
depend on https://github.com/foundriesio/ci-scripts/pull/256 so it should be merged first.

ricardosalveti:
> depend on #256 so it should be merged first.

Please update the pr!

quaresmajose:
> > depend on #256 so it should be merged first.
> 
> Please update the pr!

done

MrCry0 (approved):
LGTM

MrCry0 (approved):

helpers.sh
```diff
@@ -94,7 +94,7 @@ function set_base_lmp_version {
 	# 3: Find our base LMP version based on the HEAD
 	export LMP_VERSION=$(git describe --tags --abbrev=0 HEAD)
 	export LMP_VERSION_CACHE="$LMP_VERSION"
-	if [[ "${H_PROJECT}" == "lmp" ]] ; then
+	if [[ "${H_PROJECT}" == "lmp" ]] || [ -v LMP_VERSION_CACHE_DEV ] ; then
```
ricardosalveti, line 97: Can you explain how LMP_VERSION_CACHE_DEV would be used in this context?

I guess this is how you would have a factory using the dev cache?

quaresmajose: right, is to use the dev cache in the factory
```
The variable LMP_VERSION_CACHE_DEV is to use the development version of the cache:

# factory-config.yml

  params:
    IMAGE: lmp-factory-image
    SSTATE_CACHE_MIRROR: ""
    LMP_VERSION_CACHE_DEV: ""
```
I have already be using this for some time in some of my factories, the `lmp-ci-testing-next` that uses the latest next as well as on the `reproducible` that uses the latest stable.

quaresmajose: The `setup-environment-internal` can handle the `LMP_VERSION_CACHE_DEV`  but only when the `LMP_VERSION_CACHE` is not defined so for the ci build I add to add it too

PR #245, opened 2023-01-12
helpers: re-run repo init on loop to avoid ci failures

quaresmajose:
Sometimes repo init fails sporadically and this will abort the ci job.
To avoid such situation we can try again to run the command with a random sleep in between.

```
|== 2023-01-11 21:29:51 Running: repo init --repo-rev=v2.29.4 --no-clone-bundle -u file:///repo/.git -b baaff135ea1b0d04f6657944d755a14c1f772d72 |  repo: error: "git" failed with exit status 128
|    cwd: /srv/oe/.repo/repo
|    cmd: ['git', 'fetch', '--quiet', 'origin', '+refs/heads/*:refs/remotes/origin/*', '+refs/tags/*:refs/tags/*']
|    stderr:
|    >> fatal: unable to access 'https://gerrit.googlesource.com/git-repo/': gnutls_handshake() failed: The TLS connection was non-properly terminated.
|  fatal: cloning the git-repo repository failed, will remove '.repo/repo'
|  Downloading Repo source from https://gerrit.googlesource.com/git-repo
|--
|Script completed with error(s)
```

### DIFF
--- a/helpers.sh
+++ b/helpers.sh
@@ -54,7 +54,13 @@ function repo_sync {
 		status "Adding git config extraheader for $domain"
 		git config --global http.https://${domain}.extraheader "$(cat /secrets/git.http.extraheader)"
 	fi
-	run repo init --repo-rev=v2.29.4 --no-clone-bundle -u $* ${REPO_INIT_OVERRIDES}
+	for i in $(seq 4); do
+		run repo init --repo-rev=v2.29.4 --no-clone-bundle -u $* ${REPO_INIT_OVERRIDES} && break
+		status "repo init failed with error $?"
+		[ $i -eq 4 ] && exit 1
+		status "sleeping and trying again"
+		sleep $(($i*2))
+	done
 	for i in $(seq 4); do
 		run timeout 4m repo sync && break
 		if [ $? -eq 124 ] ; then


helpers.sh
```diff
@@ -54,7 +54,13 @@ function repo_sync {
 		status "Adding git config extraheader for $domain"
 		git config --global http.https://${domain}.extraheader "$(cat /secrets/git.http.extraheader)"
 	fi
-	run repo init --repo-rev=v2.29.4 --no-clone-bundle -u $* ${REPO_INIT_OVERRIDES}
+	for i in $(seq 4); do
+		run repo init --repo-rev=v2.29.4 --no-clone-bundle -u $* ${REPO_INIT_OVERRIDES} && break
+		run sleep $(($RANDOM%20+10))
+		if [ $i -eq 4 ] ; then
+			exit 1
```
ricardosalveti, line 61: Can you just add a print with a message saying why it is aborting the execution?

quaresmajose: The trace back is printed as I show on the commit message, I can print the return code for now and text time this happen I add the message descriptions of the return code.

quaresmajose: at this moment I don't know the reason and only the python trace back.

doanac, line 61: I don't think we need to use the "run" helper when calling sleep. We also don't need to sleep when $i=4. So maybe more like:
```
                run repo init --repo-rev=v2.29.4 --no-clone-bundle -u $* ${REPO_INIT_OVERRIDES} && break
		[ $i -eq 4 ] && exit 1
                status "repo init failed, sleeping and trying again"
                sleep $(($RANDOM%20+10))
```
Also - does this mean we could sleep up to 30s?  Maybe sleep more like `i*2` and you could add some small millesecond jitter if you are really excied.

quaresmajose: fixed

helpers.sh
```diff
@@ -54,7 +54,13 @@ function repo_sync {
 		status "Adding git config extraheader for $domain"
 		git config --global http.https://${domain}.extraheader "$(cat /secrets/git.http.extraheader)"
 	fi
-	run repo init --repo-rev=v2.29.4 --no-clone-bundle -u $* ${REPO_INIT_OVERRIDES}
+	for i in $(seq 4); do
+		run repo init --repo-rev=v2.29.4 --no-clone-bundle -u $* ${REPO_INIT_OVERRIDES} && break
+		status "repo init failed with error $?"
+		[ $i -eq 4 ] && exit 1
```
MrCry0, line 60: why don't return an actual code $?
After refactoring, we have it unchanged here since `run repo init`

quaresmajose: > why don't return an actual code $?

I can change it to exit with the actual code `$?` but there some other exits around on this script so it will not be useful for interpret by the caller (I think is jobserver)

> After refactoring, we have it unchanged here from run repo init

I didn't understand this part very well, can you say it another way please

MrCry0: > I can change it to exit with the actual code `$?` but there some other exits around on this script so it will not be useful for interpret by the caller (I think is observer)

It is a usual approach to return an actual error code of a running command.  The caller may handle return codes in different ways, if not now, so in the future. Also, it generalizes code in this function (see line 73).

> I didn't understand this part very well, can you say it another way please

I mean, as now `run sleep` is not right after `run repo init`, we have an actual `$?` from `run repo init` when calling `exit`.

doanac (approved):
Looks fine to me. I'm indifferent about capturing the repo-init error code. Feel free to merge if you are happy and have tested it in your factory.

helpers.sh
ricardosalveti: I think with the status message and the error code we're good now, since in the end the job will be aborted.

ricardosalveti (approved):
LGTM, fine to merge if tested.

MrCry0 (approved):
LGTM

quaresmajose:
Tested in my factory and ready to be merged.

doanac:
Today is a US holiday, so I'll hold off merging until I'm around .... just in case.

quaresmajose:
No problem.

PR #173, opened 2022-03-30
lmp/build: split setscene from main bitbake call

ricardosalveti:
Failures during setscene are not critical for the build as bitbake can
later rerun those tasks without cache, so separate into a dedicated call
and in a way that won't abort the job execution.

Signed-off-by: Ricardo Salveti <ricardo@foundries.io>

PR #252, opened 2023-02-02
lmp/bb-build: since lmp v87 the setscene-only is not need anymore

quaresmajose:
If we have failing setscene tasks is because our sstate cache is broken and needs to be fixed. When this happens bitbake can take care of the of the failed setscene tasks running again the real task.

Running bitbake with '--setscene-only' doesn't help in this case because when the setscene task fails, the new one stored on the sstate-cache will only be updated after running the real task.
To do that bitbake needs to run without '--setscene-only', that is what is done on the on the second bitbake call, otherwise bitbake will get the setscene task again from the sstate-cache.

This also makes more clear the sstate-cache usage in the CI because currently is not possible to see it in a clear way because in the first bitbake call the sstate-cache reuse is 100% and on the second call is 0%.

The main reason to have this is to avoid build failure when the setcene task fails but these issue is been fixed with [1] and backported to kirkstone with [2]

[1] - https://git.yoctoproject.org/poky/commit/meta/classes/sstate.bbclass?id=77f08932f4bc927b506130aa2dde4936a379ff1f
[2] - https://git.yoctoproject.org/poky/commit/meta/classes/sstate.bbclass?h=kirkstone&id=77f08932f4bc927b506130aa2dde4936a379ff1f

### DIFF
--- a/lmp/bb-build.sh
+++ b/lmp/bb-build.sh
@@ -15,8 +15,11 @@ bitbake -p
 bitbake -e > ${archive}/bitbake_global_env.txt
 bitbake -e ${IMAGE} > ${archive}/bitbake_image_env.txt
 
-# Setscene (cache), failures not critical
-bitbake --setscene-only ${IMAGE} || true
+# Before LmP version 87 (first release with OE-core kirkstone),
+# this is need to avoid build failures that can recover in the next steps
+if [ "$LMP_VERSION" -lt "87" ]; then
+    bitbake --setscene-only ${IMAGE} || true
+fi
 
 if [ "$BUILD_SDK" == "1" ] && [ "${DISTRO}" != "lmp-mfgtool" ]; then
     bitbake -D ${BITBAKE_EXTRA_ARGS} ${IMAGE} -c populate_sdk


quaresmajose:
more references for previous discussion on https://github.com/foundriesio/ci-scripts/pull/190 and https://github.com/foundriesio/ci-scripts/pull/173

ricardosalveti:
Nice, we can stage this once we know every customer we have is based on a recent enough kirkstone (which might take a while).

If needed we can have an if version < XX for based on the lmp version, and decide what to do, up to you.

quaresmajose:
> If needed we can have an if version < XX for based on the lmp version, and decide what to do, up to you.

Good ideia! I will add this `if version < XX` because I think this PR is very useful to have a better idea about the sstate-cache usage.

quaresmajose:
I added your suggestion to be enabled only before kiskstone, lmp v87.

lmp/bb-build.sh
```diff
@@ -15,8 +15,11 @@ bitbake -p
 bitbake -e > ${archive}/bitbake_global_env.txt
 bitbake -e ${IMAGE} > ${archive}/bitbake_image_env.txt
 
-# Setscene (cache), failures not critical
-bitbake --setscene-only ${IMAGE} || true
+# Before kiskstone (lmp v87) this is need to avoid build failures
```
ricardosalveti, line 18: Typo: s/kiskstone/kirkstone/g, this is needed

quaresmajose: I corrected it and emphasized the lmp version, which is what interests us

ricardosalveti (approved):
LGTM, lets merge and test on monday :-)

MrCry0 (approved):
LGTM. If you don't mind, could you update the title of the PR as well?

quaresmajose:
> LGTM. If you don't mind, could you update the title of the PR as well?

done

quaresmajose:
Tested in one of my factories and it works as expected: v86 is build in two steps and since them in a single step.

PR #253, opened 2023-02-02
tuf: Increase the TUF targets default expiration period

mike-sul:
Set the default CI targets validity period to 1 year.


doanac (approved):
looks good. not sure if there's somewhere we should document this change?

mike-sul:
@doanac @vkhoroz So, I am merging this.

PR #256, opened 2023-02-17
lmp/bb-config: fix zero length SSTATE_CACHE_MIRROR directory test

quaresmajose:
When the length of SSTATE_CACHE_MIRROR is zero the check is not evaluated correctly.

Can be tested with:
```
T=$(mktemp)
cat << 'EOF' > $T
echo -
if [ -d "$1" ]; then echo " OK "; fi
if [ -d $1   ]; then echo "FAIL"; fi
EOF
sh $T /tmp
sh $T ""
sh $T 
```


MrCry0 (approved):
LGTM

ricardosalveti (approved):
LGTM

doanac (approved):

angolini (approved):

PR #257, opened 2023-03-13
lmp/bb-config: use the full path for the bb_logconfig.json

quaresmajose:
Some tools like oe-selftest requires the full path as it runs on a relocated build path.

### DIFF
--- a/lmp/bb-config.sh
+++ b/lmp/bb-config.sh
@@ -137,7 +137,7 @@ DOCKER_MAX_DOWNLOAD_ATTEMPTS = "${DOCKER_MAX_DOWNLOAD_ATTEMPTS}"
 MFGTOOL_FLASH_IMAGE = "${MFGTOOL_FLASH_IMAGE}"
 
 # Bitbake custom logconfig
-BB_LOGCONFIG = "bb_logconfig.json"
+BB_LOGCONFIG = "${PWD}/bb_logconfig.json"
 EOFEOF
 
 # Configure path for the debug/warning logs


MrCry0 (approved):

igoropaniuk (approved):
LGTM

lmp/bb-config.sh
```diff
@@ -137,11 +137,11 @@ DOCKER_MAX_DOWNLOAD_ATTEMPTS = "${DOCKER_MAX_DOWNLOAD_ATTEMPTS}"
 MFGTOOL_FLASH_IMAGE = "${MFGTOOL_FLASH_IMAGE}"
 
 # Bitbake custom logconfig
-BB_LOGCONFIG = "bb_logconfig.json"
+BB_LOGCONFIG = "${BUILDDIR}/bb_logconfig.json"
```
ricardosalveti, line 140: Is BUILDDIR set and available by the time this line gets called?

It is not set as part of this script, but I assume it is being defined after "source setup-environment build".

ricardosalveti:
```
setup-environment-internal:    if [ -n "$BUILDDIR" ]; then
setup-environment-internal:        export BUILDDIR
```

That's probably where it is getting set.

Did you test this change on a factory run?

quaresmajose: I have tested it only in context of a bigger change to run the `oe-selftest` but will test it again (only this single commit) and come back with the results.

quaresmajose: Test on my factory and the factory log working as before. The `BUILDDIR` is also the `PWD` when this script runs and because what I need is the full path I can change it to use the `PWD` and make it not dependent of the `setup-environment-internal`

ricardosalveti: Sure, I guess PWD is better as at least it is easier to understand while reading the code.

quaresmajose: Changed to PWD but will run another test will oe-seltest too.

quaresmajose: tested with oe-seltest too

ricardosalveti (approved):
LGTM

PR #259, opened 2023-03-30
Add support to run the oe-selftest on the factory

quaresmajose:
The variable OS_SELFTEST will contain the oe-selftest arguments.
When the variable is defined on the factory-config.yml,
the oe-selftest will run with the arguments specified.

```
oe-selftest $OS_SELFTEST --newbuilddir $PWD
```

The results is also stored on the artifact folder
${archive}/selftest for post analysis.

So the following factory configuration will run the LmP reproducible test.

```
lmp:
  params:
    OS_SELFTEST: "--run-tests lmp_reproducible.ReproducibleTests.test_reproducible_builds"
```

### DIFF
--- a/lmp/bb-selftest.sh
+++ b/lmp/bb-selftest.sh
@@ -2,6 +2,11 @@
 
 source setup-environment build
 
+# The oe-selftest reproducible is build in two step (A and B) and sharing
+# the same deploy dir will cause some colisions when creating the packages
+# so use the default bitbake settings that is inside the TMP dir
+sed -i -e 's/^DEPLOY_DIR/#&/' conf/site.conf
+
 cat << EOFEOF >> conf/auto.conf
 
 # Disable non compatible classes for oe-selftest
--- a/lmp/bb-selftest.sh
+++ b/lmp/bb-selftest.sh
@@ -1,0 +1,17 @@
+#!/bin/bash -e
+
+source setup-environment build
+
+cat << EOFEOF >> conf/auto.conf
+
+# Disable non compatible classes for oe-selftest
+INHERIT:remove = "buildhistory"
+INHERIT:remove = "rm_work"
+INHERIT:remove = "create-spdx"
+INHERIT:remove = "archiver"
+EOFEOF
+
+# Save the reproducible test results
+export OEQA_DEBUGGING_SAVED_OUTPUT="${archive}/selftest"
+
+oe-selftest $OS_SELFTEST --newbuilddir $PWD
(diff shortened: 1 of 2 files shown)


lmp/build.sh
```diff
@@ -62,6 +62,11 @@ touch ${archive}/bitbake_debug.log ${archive}/bitbake_warning.log ${archive}/bit
 touch ${archive}/bitbake_global_env.txt ${archive}/bitbake_image_env.txt && chown builder ${archive}/bitbake_*_env.txt
 touch ${archive}/app-preload.log && chown builder ${archive}/app-preload.log
 touch ${archive}/tuf-root-fetch.log && chown builder ${archive}/tuf-root-fetch.log
+if [ ! -z "$OS_SELFTEST" ]; then
```
ricardosalveti, line 65: ! -z can be replaced by -n

lmp/build.sh
```diff
@@ -62,6 +62,11 @@ touch ${archive}/bitbake_debug.log ${archive}/bitbake_warning.log ${archive}/bit
...
+	mkdir ${archive}/selftest && chown builder ${archive}/selftest
+	su builder -c $HERE/bb-selftest.sh
+	exit
```
ricardosalveti, line 68: Exit will abort the script and the resulted data won't be uploaded, is that what you want here?

quaresmajose: No, I like to upload the artifacts but also return the exit code of the last command.
Perhaps I need to save the return code of the `$HERE/bb-selftest.sh` and run the `lmp/build.sh` until the end.

quaresmajose: I think the reproducible test is the only one that produce artifacts and all the others just write log on the terminal.
but I need to confirm this by running some other tests.

lmp/build.sh
quaresmajose: fixed

mike-sul:
@ricardosalveti @quaresmajose Do we need to merge it?

lmp/build.sh
quaresmajose: @ricardosalveti I have confirmed in my `reproducible` factory and there is a `selftest` folder on the build artifacts. even when the build fails.
I think it works because the folder is already on `${archive}/selftest` so don't need to be move like some others.

lmp/bb-selftest.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash -e
+
+source setup-environment build
+
+cat << EOFEOF >> conf/auto.conf
+
+# Disable problematic classes for oe-selftest
```
angolini, line 12: maybe replace problematic by non compatible

quaresmajose: I think I use problematic because is what bitbake says when they are enabled. Fixed

quaresmajose:
> I like the PR, however, my concern is how to properly explain how to use the results. Have you thought about how to document it already?

The `oe-selftest` is very big and run a bunch of different things. This PR started to run the `reproducible` test and is that's what I always tested, however I noticed that this could be generic and work for all tests that runs with the `oe-seftest` so I change it to be generic.

Anyway the `reproducible` test produces some artifacts to be analyzed when the test fails, in short it's an html page and the packages that failed.
One piece is still in my factory so is not possible to run `reproducible` test without it.
I need to move it to the meta-lmp layer to be used as reference and maybe after that I can write something (bog/doc) to explain how to use/run the `reproducible` test.

angolini:
It's a good idea to document how to run/enable. But I also think it could be good to define the test scope as well.

MrCry0 (approved):

ricardosalveti (approved):
LGTM

PR #260, opened 2023-04-05
Fetch apps unconditionally

mike-sul:
Fetch and store Restorable Apps regardless whether the assemble run succeeds with Apps preloading or not.
Therefore, a user can download an apps tarball for offline update even if app preloading fails, e.g. a WIC image is not supported.

### DIFF
--- a/assemble.py
+++ b/assemble.py
@@ -171,15 +171,15 @@ def _mk_parent_dir(path: str):
 
 def copy_compose_apps_to_wic(target: FactoryClient.Target, fetch_dir: str, wic_image: str, token: str,
                              apps_shortlist: list, progress: Progress):
-    p = Progress(4, progress)
+    p = Progress(3, progress)
     apps_fetcher = TargetAppsFetcher(token, fetch_dir)
     apps_fetcher.fetch_target(target, shortlist=apps_shortlist, force=True)
     p.tick()
     apps_size_b = apps_fetcher.get_target_apps_size(target)
-    p.tick()
 
     logger.info('Compose Apps require extra {} bytes of storage'.format(apps_size_b))
     with WicImage(wic_image, apps_size_b) as wic_image:
+        p.tick()
         if os.path.exists(wic_image.docker_data_root):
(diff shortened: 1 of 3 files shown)


assemble.py
```diff
@@ -295,8 +302,23 @@ def get_args():
...
+            release_info = factory_client.get_target_release_info(target)
+            if release_info.lmp_version > 0:
+                logger.info(
+                    f"Target's LmP version: {release_info.lmp_version}, yocto version: {release_info.yocto_version}")
+
+            if args.app_type == 'restorable' or (not args.app_type and release_info.lmp_version > 84):
+                logger.info('Fetching Restorable Apps...')
+                apps_desc = fetch_restorable_apps(target, apps_root_dir, args.app_shortlist, args.token)
+                fetched_apps[target.name] = (apps_desc, os.path.join(args.out_image_dir, target.tags[0]))
+            p.tick(complete=True)
```
doanac: Not sure `complete=True` is correct here when len(targets) > 1?

doanac: n/m you fix this later on.

doanac (commented):
just one small nit about the error that gets logged for an http failure.

factory_client.py
```diff
@@ -163,18 +162,22 @@ def get_target_system_image(self, target: Target, out_dir: str, progress: Progre
...
+        image_machine = target['custom']['hardwareIds'][0]
+        os_release_url = os.path.join(base_url, 'runs', image_machine, 'os-release')
+        release_info = self.Release(0, '')
+        try:
+            release_resp = requests.get(os_release_url, headers=self._auth_headers)
+            release_resp.raise_for_status()
+            release_info = self.Release.parse(dict([line.split('=') for line in release_resp.content.decode().splitlines()]))
+        except Exception as exc:
+                logger.error(f"Failed to get info about Target's LmP release; target: {target.name},"
+                             " err: " + str(exc))
```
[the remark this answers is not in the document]
mike-sul: Replaced `raise_for_status` with `http_get` (force push).

doanac (approved):

PR #261, opened 2023-04-07
Preload to ext4

mike-sul:
Fallback to a `.ext4` system image usage if `.wic` is not found for apps preloading.

### DIFF
--- a/assemble.py
+++ b/assemble.py
@@ -9,6 +9,8 @@ import json
 import argparse
 import logging
 import shutil
+
+import requests
 from math import ceil
 from time import sleep
 from typing import NamedTuple
@@ -342,7 +344,24 @@ if __name__ == '__main__':
                 preload_progress.tick(complete=True)
                 continue
 
-            image = factory_client.get_target_system_image(target, args.out_image_dir, preload_progress)
+            try:
+                image = factory_client.get_target_system_image(target, args.out_image_dir,
+                                                               preload_progress)
+            except requests.HTTPError as exc:
+                if exc.response.status_code == 404:
+                    # try to download `ota-ext4` image
+                    logger.info("Target's system image in `.wic` format was not found,"
+                                " trying to get an `.ota-ext4` image;"
+                                " not found path: " + exc.response.url)
+                    image = \
+                        factory_client.get_target_system_image(target, args.out_image_dir,
+                                                               preload_progress, format=".ota-ext4")
+                else:
+                    raise requests.exceptions.\
+                        HTTPError('Failed to get {}: HTTP_{}\n{}'.format(exc.response.url,
+                                                                         exc.response.status_code,
+                                                                         exc.response.text))
+
             if target.name in fetched_apps:
                 logger.info('Preloading Restorable Apps...')
                 copy_restorable_apps_to_wic(target, image, fetched_apps[target.name][0], preload_progress)
--- a/assemble.py
+++ b/assemble.py
@@ -70,17 +70,18 @@ def losetup(path: str) -> str:
     raise RuntimeError(f'Unable to find loop device for {path}')
 
 
-class WicImage:
+class ImageVolume:
     ComposeAppsRootDir = 'ostree/deploy/lmp/var/sota/compose-apps/'
     DockerDataRootDir = 'ostree/deploy/lmp/var/lib/docker/'
     RestorableAppsRoot = 'ostree/deploy/lmp/var/sota/reset-apps'
     InstalledTargetFile = 'ostree/deploy/lmp/var/sota/import/installed_versions'
 
-    def __init__(self, wic_image_path: str, increase_bytes=None, extra_space=0.2):
-        self._path = wic_image_path
-        self._mnt_dir = os.path.join('/mnt', 'wic_image_rootfs')
+    def __init__(self, image_path: str, increase_bytes=None, extra_space=0.2):
+        self._path = image_path
+        self._mnt_dir = os.path.join('/mnt', 'image_rootfs')
         self._installer_mount = None
-        self._last_part = 2
+        self._part_numb, self._gpt = self._get_last_part(self._path)
+        logger.info(f'Detected last partition is {self._part_numb}, going to preload apps into it')
         self._resized_image = False
         if increase_bytes:
             self._resize_wic_file(increase_bytes, extra_space)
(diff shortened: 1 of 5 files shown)


factory_client.py
```diff
@@ -149,7 +149,8 @@ def get_target_system_image(self, target: Target, out_dir: str, progress: Progre
             logger.info('Downloading Target system image...; Target: {}, image: {}'
                         .format(target.name, image_filename))
 
-            image_resp = http_get(image_url, headers=self._auth_headers)
+            image_resp = requests.get(image_url, headers=self._auth_headers)
+            image_resp.raise_for_status()
```
doanac, line 159: Can you share how the program output will differ. The way it was looks like you get roughly the same thing:
 https://github.com/foundriesio/ci-scripts/blob/ff4e797d5fd005ddb37bcea3a6a8efcc0364f168/helpers.py#L151

assemble.py
```diff
@@ -47,20 +47,26 @@ def losetup(path: str) -> str:
...
         remount_dev()
+        # get the next available loop device
+        loop_device = cmd('losetup', '-f', capture=True).decode().rstrip()
         cmd('losetup', '-P', '-f', path)
 
     # The -P in losetup scans for partitions and will create entries like:
     # /dev/loopXp1. Since these are new /dev entries, we have to remount /dev
     remount_dev()
 
+    # make sure that the most recently created loop device represents a given system image (`path`)
```
doanac, line 66: I'm assuming you hit this in local development and its not really something that can happen in actual CI, right?

assemble.py
```diff
@@ -47,20 +47,26 @@ def losetup(path: str) -> str:
     # up, thus the "No such file or directory". When this happens, we need to
     # remount /dev in order to see the new device.
     try:
+        # get the next available loop device
+        loop_device = cmd('losetup', '-f', capture=True).decode().rstrip()
```
doanac, line 53: Could you change the line below to be `cmd('losetup', '-P', loop_device, path)`?

assemble.py
```diff
@@ -93,7 +98,8 @@ def __init__(self, wic_image_path: str, increase_bytes=None, extra_space=0.2):
 
     def __enter__(self):
         self._loop_device = losetup(self._path)
-        self._wic_device = self._loop_device + 'p' + str(self._last_part)
+        self._wic_device = \
+            self._loop_device if self._last_part == 1 else f"{self._loop_device}p{self._last_part}"
```
doanac, line 104: So ext4 does `self._loop_device` and wic appens a "p2"? I don't want to over-engineer this, but that's really subtle and not bound to the "SupportedFormatsToPartitionNumb". Maybe we need some type of "FileSystemFormat" object like:
```
class WicImage:
    @property
    def partition_num(self):
        return 2
    def get_device_name(self, loop_device):
        return loop_device + 'p' + str(self._last_part)

class Ext4Image:
    @property
    def partition_num(self):
        return 1
    def get_device_name(self, loop_device):
        return loop_device
```

assemble.py
```diff
@@ -338,7 +340,21 @@ def get_args():
...
+                if exc.response.status_code == 404:
+                    # try to download `ota-ext4` image
+                    logger.info("Target's system image in `.wic` format was not found,"
+                                " trying to get an `ext4` image;"
+                                " not found path: " + exc.request.path)
+                    image_file_path = \
+                        factory_client.get_target_system_image(target, args.out_image_dir,
+                                                               preload_progress, format=".ota-ext4")
+                else:
+                    raise
```
doanac, line 356: I had envisioned making this a factory config setting, but I think this actually make sense and is easier for the user - so long as we explain it.

assemble.py
```diff
@@ -293,40 +313,71 @@ def get_args():
...
-            apps_dir = None
-            if args.app_type == 'restorable' or (not args.app_type and release_info.lmp_version > 84):
+            try:
+                image_file_path = factory_client.get_target_system_image(target, args.out_image_dir,
+                                                                         preload_progress)
+            except requests.HTTPError as exc:
+                if exc.response.status_code == 404:
+                    # try to download `ota-ext4` image
+                    logger.info("Target's system image in `.wic` format was not found,"
+                                " trying to get an `ext4` image;"
```
ricardosalveti, line 350: And probably better to say it is looking for an ota-ext4 image, as a target can also generate a "normal" ext4 image in yocto (without ostree support).

factory_client.py
mike-sul: I need a status code of a response to catch `404`. `raise_for_status` adds a response object to the HTTPError context while `http_get` doesn't do it.
An alternative to this, could be altering the `http_get` implementation so it adds a response object or just a status code to an exception context.
I am fine with either options.

assemble.py
mike-sul: Yes, I hit it locally. I think it may happen in actual CI though. E.G. if a run is cancelled or a run container is killed, so `losetup -d` is never called. As result, the next run on that host/worker can use a loop device that mounts the previous run's image.

factory_client.py
doanac: ah - i see now. the caller of this function is catching the error.

assemble.py
mike-sul: I think so.

assemble.py
mike-sul: Initially I thought about some OOP too. Then I realized that the implementation is not format specific. A device name/path depends on number of partitions inside of an input image file regardless whether it's WIC or ext4 (ext4 may have two partitions two.)

I actually think that we should rename `WicImage` class to something like `Image` and pass a target partition number as an input parameter (`Image(image_path, part_numb, ...)`). A client of `Image` class knows which of the image partitions should be preloaded, so `Image` becomes generic and agnostic of an use-case.

assemble.py
mike-sul: Yep, fallback seems easier atm and fulfill the requirement.

assemble.py
mike-sul: done

assemble.py
mike-sul: Done

assemble.py
mike-sul: Done

doanac (approved):

mike-sul:
@doanac I think that's all what I wanted to do in a scope of the given PR.

doanac:
> @doanac I think that's all what I wanted to do in a scope of the given PR.

Yep - I've already approved. ship it

PR #262, opened 2023-04-12
assemble: Skip compose Apps preloading if not needed

mike-sul:
Don't preload compose Apps if restorable Apps are preloaded and LmP version is >= than 87.
LmP >= 87 includes the apps early startup scripts for restorable apps what makes compose apps preloading redundant/not needed.

The early startup script includes "image registration" at the docker store before running `dockerd` so the later becomes aware about images copied to the store by `skopeo copy` command.

### DIFF
--- a/assemble.py
+++ b/assemble.py
@@ -329,7 +329,7 @@ if __name__ == '__main__':
             if release_info.lmp_version > 0:
                 logger.info(
                     f"Target's LmP version: {release_info.lmp_version}, yocto version: {release_info.yocto_version}")
-
+            target.lmp_version = release_info.lmp_version
             if args.app_type == 'restorable' or (not args.app_type and release_info.lmp_version > 84):
                 logger.info('Fetching Restorable Apps...')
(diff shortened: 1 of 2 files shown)


mike-sul:
I verified it e2e, just restorable apps preloading + early startup works just fine.

assemble.py
```diff
@@ -366,8 +366,13 @@ def get_args():
                 logger.info('Preloading Restorable Apps...')
                 copy_restorable_apps_to_wic(target, image, fetched_apps[target.name][0], preload_progress)
 
-            logger.info('Preloading Compose Apps...')
-            copy_compose_apps_to_wic(target, args.fetch_dir + "/compose", image, args.token, args.app_shortlist, preload_progress)
+            if target.name not in fetched_apps or target.lmp_version < 87:
```
doanac, line 369: am i missing something subtle about why you don't use `release_info.lmp_version` here?

mike-sul: There is no `release_info` in this scope/loop.

doanac (approved):

PR #269, opened 2023-05-22
apps: Create a helper for docker-login

doanac:
This consolidates the docker-login logic to a single helper.

This also changes the container build process only call docker-login once against hub.fio which is all that's needed. Calling multiple times has no effect. We are also long past the time you could run this script with "noauth", so I simplified the logic

### DIFF
--- a/apps/build.sh
+++ b/apps/build.sh
@@ -55,9 +55,7 @@ if [ -f $pbc ] ; then
   . $pbc
 fi
 
-if [ -f /secrets/container-registries ] ; then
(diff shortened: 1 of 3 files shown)


doanac:
@mike-sul - this is part of a bigger "meds" branch i have. I'm going to try and start upstreaming changes in smaller chucks where possible. Testing here:

 https://app.foundries.io/factories/andy-corp/targets/12/

mike-sul (approved):

PR #274, opened 2023-06-05
Updates to support configurable URLs to our backend

doanac:
These changes allow us to run CI jobs against services not under foundries.io.

### DIFF
--- a/apps/build.sh
+++ b/apps/build.sh
@@ -22,6 +22,8 @@ BUILDKIT_VERSION="${BUILDKIT_VERSION-v0.10.3}"
 MANIFEST_PLATFORMS_DEFAULT="${MANIFEST_PLATFORMS_DEFAULT-linux/amd64,linux/arm,linux/arm64}"
 status Default container platforms will be: $MANIFEST_PLATFORMS_DEFAULT
 
+load_extra_certs
+
 if [ -f /secrets/docker_host_config.json ] ; then
 	mkdir -p $HOME/.docker
 	cp /secrets/docker_host_config.json $HOME/.docker/config.json
--- a/apps/build.sh
+++ b/apps/build.sh
@@ -106,7 +106,7 @@ for x in $IMAGES ; do
 	# allow the docker-build.conf to override our manifest platforms
 	MANIFEST_PLATFORMS="${MANIFEST_PLATFORMS-${MANIFEST_PLATFORMS_DEFAULT}}"
 
-	ct_base="hub.foundries.io/${FACTORY}/$x"
+	ct_base="${hub_fio}/${FACTORY}/$x"
 
 	docker_cmd="$docker_build -t ${ct_base}:$TAG-$ARCH -t ${ct_base}:$LATEST-$ARCH --force-rm"
 	if [ -z "$NOCACHE" ] ; then
@@ -155,12 +155,12 @@ for x in $IMAGES ; do
 	for t in $(eval echo "\$$var") ; do
 		status "Handling manifest logic for $var"
 
-		tmp=$HOME/.docker/manifests/hub.foundries.io_${FACTORY}_${x}-${H_BUILD}_${TAG}
-		cp ${tmp}/hub.foundries.io_${FACTORY}_${x}-${TAG}-${ARCH} ${tmp}/hub.foundries.io_${FACTORY}_${x}-${TAG}-${t}
+		tmp=$HOME/.docker/manifests/${hub_fio}_${FACTORY}_${x}-${H_BUILD}_${TAG}
+		cp ${tmp}/${hub_fio}_${FACTORY}_${x}-${TAG}-${ARCH} ${tmp}/${hub_fio}_${FACTORY}_${x}-${TAG}-${t}
 		run docker manifest annotate ${ct_base}:${H_BUILD}_${TAG} ${ct_base}:${TAG}-$t --arch $t
 
-		tmp=$HOME/.docker/manifests/hub.foundries.io_${FACTORY}_${x}-${LATEST}
-		cp ${tmp}/hub.foundries.io_${FACTORY}_${x}-${TAG}-${ARCH} ${tmp}/hub.foundries.io_${FACTORY}_${x}-${TAG}-${t}
+		tmp=$HOME/.docker/manifests/${hub_fio}_${FACTORY}_${x}-${LATEST}
+		cp ${tmp}/${hub_fio}_${FACTORY}_${x}-${TAG}-${ARCH} ${tmp}/${hub_fio}_${FACTORY}_${x}-${TAG}-${t}
 		run docker manifest annotate ${ct_base}:${LATEST} ${ct_base}:${TAG}-$t --arch $t
 	done
 
--- a/apps/build.sh
+++ b/apps/build.sh
@@ -23,7 +23,6 @@ MANIFEST_PLATFORMS_DEFAULT="${MANIFEST_PLATFORMS_DEFAULT-linux/amd64,linux/arm,l
 status Default container platforms will be: $MANIFEST_PLATFORMS_DEFAULT
 
 load_extra_certs
-
 if [ -f /secrets/docker_host_config.json ] ; then
 	mkdir -p $HOME/.docker
 	cp /secrets/docker_host_config.json $HOME/.docker/config.json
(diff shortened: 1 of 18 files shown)


helpers.sh
```diff
@@ -26,6 +26,13 @@ function require_params {
 	done
 }
 
+function load_extra_certs {
```
mike-sul, line 35: Maybe we should do it the container image during its building?

apps/build.sh
```diff
@@ -106,7 +106,7 @@ for x in $IMAGES ; do
 	# allow the docker-build.conf to override our manifest platforms
 	MANIFEST_PLATFORMS="${MANIFEST_PLATFORMS-${MANIFEST_PLATFORMS_DEFAULT}}"
 
-	ct_base="hub.foundries.io/${FACTORY}/$x"
+	ct_base="${hub_fio}/${FACTORY}/$x"
```
mike-sul, line 118: Where `hub_fio` is defined? In the login function. Maybe we should introduce some env. var for it.

apps/build.sh
```diff
@@ -58,6 +57,16 @@ if [ -f $pbc ] ; then
 fi
 
 docker_login
+buildx_create_args="--driver-opt image=moby/buildkit:${BUILDKIT_VERSION} --use"
+if [ -d /usr/local/share/ca-certificates ] ; then
+	# We need to pass these certs into the buildkit container
+	bkconf="/buildkit.toml"
+	cat >${bkconf} <<EOF
+[registry."${hub_fio}"]
+ca=["/etc/ssl/certs/ca-certificates.crt"]
```
mike-sul, line 66: I guess it cannot be specified via a CLI param?

apps/build.sh
```diff
@@ -58,6 +57,16 @@ if [ -f $pbc ] ; then
 fi
 
 docker_login
+buildx_create_args="--driver-opt image=moby/buildkit:${BUILDKIT_VERSION} --use"
+if [ -d /usr/local/share/ca-certificates ] ; then
+	# We need to pass these certs into the buildkit container
+	bkconf="/buildkit.toml"
```
mike-sul, line 63: I suppose this file can be generating during the container image building (lmp-inage-tools).

helpers.sh
```diff
@@ -7,8 +7,12 @@ function status { echo == $(date "+%F %T") $* ; }
 
 function run { set -o pipefail; status "Running: $*"; $* 2>&1 | indent ; }
 
+function dns_base {
+	echo $H_RUN_URL | cut -d/ -f3 | sed -e 's/api.//'
+}
+
 function docker_login {
-	hub_fio=$(echo $H_RUN_URL | cut -d/ -f3 | sed -e 's/api./hub./')
+	hub_fio="hub.$(dns_base)"
```
mike-sul, line 15: Why not use `FIO_HUB_URL` here?

lmp/bb-config.sh
```diff
@@ -10,7 +10,9 @@ AKLITE_TAG="${AKLITE_TAG-promoted}"
 H_BUILD="${H_BUILD-lmp-localdev}"
 LMP_VERSION="${LMP_VERSION-unknown}"
 FACTORY="${FACTORY-lmp}"
-LMP_DEVICE_API="${LMP_DEVICE_API-https://api.foundries.io/ota/devices/}"
+LMP_DEVICE_API="${LMP_DEVICE_API-https://api.$(dns_base)/ota/devices/}"
+LMP_OAUTH_API="${LMP_OAUTH_API-https://app.$(dns_base)/oauth}"
+FIO_HUB_URL="${FIO_HUB_URL-hub.$(dns_base)}"
```
mike-sul, line 15: Maybe we should do the same for apps/publish?

mike-sul (approved):
LGTM as long as we are fine with hard-coding the domain prefixes (api, hub, etc).

apps/build.sh
doanac: the docker-login function sets this as a variable

apps/build.sh
doanac: Not according to what I read

apps/build.sh
doanac: I don't think we want a custom image per instance.

helpers.sh
doanac: it would be the other way around, but FIO_HUB_URL is set in bb-config.sh. Its only available to a yocto build. this is getting setting for container builds.

lmp/bb-config.sh
doanac: i think apps/publish handles this also?

apps/build.sh
mike-sul: Is this directory `/usr/local/share/ca-certificates` of container mapped to a host directory?

apps/build.sh
```diff
@@ -56,6 +57,16 @@ if [ -f $pbc ] ; then
...
+	cat >${bkconf} <<EOF
+[registry."${hub_fio}"]
```
mike-sul, line 65: Ah., this is an instance specific.

apps/build.sh
doanac: exactly - https://github.com/foundriesio/jobserv/commit/e7affa0eae6ae7ae8248bdb7cc7a8842f9430b44

apps/build.sh
mike-sul: `ca=["/etc/ssl/certs/ca-certificates.crt"]` but we check `if [ -d /usr/local/share/ca-certificates ]`, I mean we check different directory than use?

doanac: the `load_extra_certs` takes the certs and produces an updated `/etc/ssl/certs/ca-certificates.crt` - so this more about making the container use the same certs as this "host"

mike-sul: ok, a bit subtle.

helpers.sh
mike-sul: I understand it. My point is that defining the env. var. at the beginning of the overall script (maybe publish.sh) is less subtle then defining `hub_fio` in one of the functions and then using it at the other places.

lmp/bb-config.sh
mike-sul: I just thought about adding those env. vars to `publish.sh`.

mike-sul:
LGTM. Those comments are not important. It would be great to fix the unit tests too.

helpers.sh
doanac: we've never defined variable globally in something like helpers.sh so that multiple scripts get a value. We could do something like the `dns_base` function in helpers.sh though. Then everyone has access to it.

doanac: I've cleaned this up a little bit and just set dns_base and hub_fio as environment variables. gonna test this stuff out and then look at the unit-test

doanac:
@mike-sul - unit tests are passing and its working in meds and my andy-corp factory. Care to give this a spin before I merge?

lmp/bb-config.sh
```diff
@@ -49,6 +51,15 @@ EULA_stm32mp15disco="1"
 EULA_stm32mp1eval="1"
 EULA_stm32mp15eval="1"
 
+# meta-lmp requires this value to be a script and not a script with arguments.
+# This creates a wrapper that handles the arguments
+fetch_creds=$(mktemp --suffix=.sh)
```
mike-sul, line 56: Maybe if we set `H_RUN_URL=${H_RUN_URL}` in `conf/local.conf` then we will not need this dynamic scripting? Similar case (`GARAGE_TARGET_VERSION = "${H_BUILD}"` at #94.

doanac: you are probably right. @ricardosalveti - is mike's suggestion preferable for this?

mike-sul: I think it's not so important to be a blocker for //M.

quaresmajose: IMO we should avoid using `mktemp` in bitbake variable because it breaks the sstate cache, they generate a different values in each run. Something like `${HERE}/fetch-root-meta-helper.sh` look better to me.

This variable seems to only affect the tasks of the final target image that always runs on the CI build (do not use sstate-cache) but avoid the `mktemp` usage on variable names is good police in my opinion.

quaresmajose: The script to looks good to me, it's just the name that worried me because they are not reproducible

doanac: rebased on latest master and then fixed in force-push

PR #276, opened 2023-06-14
assemble: Fetch apps to the archive dir by default

mike-sul:
Fetch app images to the archive sub-directory (host dir mounted to the container) by default.

Currently, app images are fetched to the container `tmp` dir mapped to a host's RAM. The host's RAM capacity can be lower than the overall size of extracted/pulled App images, hence we need to pull them into some other folder.

Not mapping the container's tmp implies that the tmp is based on the host's overlay2 FS (the CI run container FS); it doesn't work too because of the "overlay2 over overlay2 issue", dockerd cannot have a data root based on overlay2 FS.

Therefore, the only feasible solution left is to fetch app images to the archive sub-directory which is mapped to a host's directory located on a regular volume.

### DIFF
--- a/assemble.py
+++ b/assemble.py
@@ -299,6 +299,7 @@ def get_args():
 if __name__ == '__main__':
     exit_code = 0
     fetched_apps = {}
+    fetch_dir = ""
     image = ""
     p = Progress(total=3)  # fetch apps, preload images, move apps to the archive dir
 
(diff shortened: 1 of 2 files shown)


assemble.py
```diff
@@ -391,5 +394,10 @@ def get_args():
         os.makedirs(dst_dir, exist_ok=True)
         cmd('tar', '-cf', os.path.join(dst_dir, target + '-apps.tar'), '-C', apps_desc.dir, '.')
 
+    # Cleanup the fetched images
+    if os.path.exists(fetch_dir):
+        logger.info(f'Removing `{fetch_dir}` directory Apps were fetched to...')
+        shutil.rmtree(fetch_dir, ignore_errors=True)
```
mike-sul, line 400: The `image` sub-dir has some odd attributes after been used by `dockerd` so we need to ignore the removal error. It doesn't impact removal anyhow, all data are removed.

PR #277, opened 2023-06-14
assemble: Fetch apps to the NFS storage by default

mike-sul:
Fetch app images to the NFS sub-directory (NFS dir mounted to the container) by default.

Currently, app images are fetched to the container `tmp` dir mapped to a host's RAM. The host's RAM capacity can be lower than the overall size of extracted/pulled App images, hence we need to pull them into some other folder.

Not mapping the container's tmp implies that the tmp is based on the host's overlay2 FS (the CI run container FS); it doesn't work too because of the "overlay2 over overlay2 issue", dockerd cannot have a data root based on overlay2 FS.

Therefore, the only feasible solution left is to fetch app images to the NFS sub-directory which is mapped to a host's directory located on a NFS volume.

### DIFF
--- a/assemble-system-image.sh
+++ b/assemble-system-image.sh
@@ -20,7 +20,7 @@ OUT_IMAGE_DIR="${OUT_IMAGE_DIR-/archive}"
 APP_SHORTLIST="${APP_SHORTLIST-""}"
 COMPOSE_APP_TYPE=${COMPOSE_APP_TYPE-""}
 # directory to preload/dump/snapshot apps images to
-FETCH_DIR="${FETCH_DIR-$(mktemp -u -d)}"
+FETCH_DIR="${FETCH_DIR-$(mktemp -u -d -p /var/cache/apps)}"
 
 require_params FACTORY OUT_IMAGE_DIR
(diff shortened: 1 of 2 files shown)


doanac:
There is not NFS in container builds. `/var/cache/bitbake/fetched-apps` will just be a directory within the container.

mike-sul:
> There is not NFS in container builds. `/var/cache/bitbake/fetched-apps` will just be a directory within the container. we need to update our git server to generate CI runs that will have "persistent volumes". then mount that into these "assemble" CI containers.

Sure, the change to repoz

```
'persistent-volumes': {
                    'bitbake': '/var/cache/bitbake',
                },
```

mike-sul:
Maybe we should just fetch to the archive https://github.com/foundriesio/ci-scripts/pull/276 as fetching to NFS may have performance impact.

doanac:
> Maybe we should just fetch to the archive #276 as fetching to NFS may have performance impact.

I think adding a persistent volume for this is probably worth the trouble.

mike-sul:
> And maybe not call it "bitbake" since it won't point at that and is different.

What about `'apps': '/var/cache/apps'`?

doanac (approved):

PR #278, opened 2023-06-19
publish: Strip spaces around machine names

mike-sul:


apps/publish.py
```diff
@@ -68,7 +68,7 @@ def get_args():
     try:
         logging.basicConfig(format='%(asctime)s %(levelname)s: %(module)s: %(message)s', level=logging.INFO)
         args = get_args()
-        machines = args.machines.split(',') if args.machines else None
+        machines = [x.strip for x in args.machines.split(',') if x] if args.machines else None
```
doanac, line 71: needs to be `x.strip()`

doanac: actually more than that. should be
```
[x.strip() for x in args.machines.split(',') if x.strip()]
```

mike-sul: fixed

doanac (approved):

PR #279, opened 2023-06-23
lmp: Move to "main" OTA tag

doanac:
Our next release will move users to having an initial factory based on a "fast first target". This target will be the LmP prebuilt release image. These images currently set their default OTA tag as "postmerge". However, the new factories we create for people will be set to "main".

This change makes the tag "main" so that `lmp-device-register` will put the device on the correct initial tag.


ricardosalveti (approved):
LGTM

PR #281, opened 2023-06-27
Teach static delta logic to work with generic URLs

doanac:


helpers.py
```diff
@@ -197,6 +197,12 @@ def generate_credential_tokens(creds_zip_out: str):
         zout.writestr('tufrepo.url', f'https://api.{base}/ota/repo/{factory}')
 
 
+def load_extra_certs():
+    if os.path.exists("/usr/local/share/ca-certificates"):
+        status("Loading extra ca ca certificates")
+        cmd("update-ca-certificates")
```
mike-sul, line 203: I suppose you verified that this command `update-ca-certificates` will work in the container used for the static deltas generation.

mike-sul (approved):

PR #286, opened 2023-07-11
lmp/build.sh: create bootchart.svg with the buildstats collected data

quaresmajose:
In LmP build we have the buildstats enabled because we use the
INHERIT += "buildstats" but the output is hard to read and understand.
The openembedded-core pybootchartgui scripts can process this logs and
generates some pretty charts.

### DIFF
--- a/lmp/bb-build.sh
+++ b/lmp/bb-build.sh
@@ -31,6 +31,12 @@ BUILDSTATS_PATH="$(bitbake-getvar --value TMPDIR | tail -n 1)/buildstats"
 if [ -d $BUILDSTATS_PATH ]; then
     # get the most recent folder
     BUILDSTATS_PATH="$(ls -td -- $BUILDSTATS_PATH/*/ | head -n 1)"
+    # we need to check that because it can't be available in old containers
+    if command -v xvfb-run >/dev/null 2>&1 ; then
+        # producing bootchart.svg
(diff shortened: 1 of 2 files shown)


lmp/bb-build.sh
```diff
@@ -26,13 +26,18 @@ if [ "$BUILD_SDK" == "1" ] && [ "${DISTRO}" != "lmp-mfgtool" ]; then
...
-    if [ -d $BUILDSTATS_PATH ]; then
-        # get the most recent folder
-        BUILDSTATS_PATH="$(ls -td -- $BUILDSTATS_PATH/*/ | head -n 1)"
+# Check if the buildstats was enabled
+BUILDSTATS_PATH="$(bitbake-getvar --value TMPDIR | tail -n 1)/buildstats"
+if [ -d $BUILDSTATS_PATH ]; then
+    # get the most recent folder
+    BUILDSTATS_PATH="$(ls -td -- $BUILDSTATS_PATH/*/ | head -n 1)"
+    # producing bootchart.svg
+    run xvfb-run ../layers/openembedded-core/scripts/pybootchartgui/pybootchartgui.py \
```
ricardosalveti, line 35: Can you add a check if xvfb is available and skip if not? Then this same logic would work when running ci-scripts on older lmp-sdks.

quaresmajose: good cacth! fixed

ricardosalveti (approved):
LGTM

PR #288, opened 2023-07-14
Handle uri with tags

mike-sul:


apps/get_layers_meta.py
```diff
@@ -51,7 +50,8 @@ def print_layer_details(layer: DockerStore.Layer):
         for img in app.images(expand_env=True):
             img_uri = img
             if img_uri.startswith("hub.foundries.io"):
-                img_uri += ":" + args.tag
+                end_pos = img_uri.rfind(":")
+                img_uri = img_uri[0:end_pos if end_pos != -1 else len(img_uri)] + ":" + args.tag
```
doanac, line 54: this is a lot of logic in one line. I think you are turning something from:
 `hub.foundries.io/foo/bar:tag` to `hub.foundries.io/foo/bar:{args.tag}` ?

Seems like it should be something like:
```
if end_pos == -1:
 img_uri += ":" args.tag
```

doanac: but if its tagged why are you replacing the tag? that seems wrong?

mike-sul: The publish CI run allows only `latest` custom tag for images built by our CI builder and replaces it with the tag that corresponds to a current CI build. Our a CI container build runner tags the built images as `docker_cmd="$docker_build -t ${ct_base}:$TAG-$ARCH -t ${ct_base}:$LATEST-$ARCH --force-rm"`.
Therefore, the current logic needs to replace any images with `:latest` tag to the proper tag `$TAG-$ARCH`.

mike-sul: Effectively, there is no reference to an image with "latest" tag in the local docker store.

doanac: k. so maybe we just need a comment - and probably an assertion that its ":latest" ?

mike-sul: I think it's worth adding comment, not sure about the assertion for "latest" though.

Currently, a build run doesn't fail if an image is not tagged "correctly" in `docker-compose.yml`. Only the publish run checks it. So, I thought about leaving this logic intact. To some extend, a build run doesn't care much about an image tag specified in `docker-compose.yml`, it just builds an image defined in `Dockerfile` and tag it according to a commit hash and architecture. And, this logic runs within a build run.

doanac: sounds good - as long as the logic is consistent.

mike-sul: Added the comment.

doanac: i can't articulate it well, but i'm worried this is going to cause some regression as the logic we have at - https://github.com/foundriesio/ci-scripts/blob/master/apps/apps_publisher.py#L48 is much more nuanced. i.e. if the image is pinned to a sha256 and not a tag (I *think* a customer is doing that). And if they point a "public" lmp container image (real edge case and they shouldn't, but its possible)

mike-sul: If the image is pinned it means that the image is not built by a given CI build run. So, the code will just print this message https://github.com/foundriesio/ci-scripts/pull/288#discussion_r1265871192 and continue its work. Then, the publish run will fail if we don't do the fallback to using the approximate app update calculation (we don't know image layers disk usage if the image is not built in the current CI, but we can get infor about layer archive/blob sizes).
That's why I did the fallback in the latest PR to `compose-publish` repo.

mike-sul: > if someone has container referenced by sha256 that may not be built in our CI.

We are discussing the case `if img_uri.startswith("hub.foundries.io"):` . Images built out of the fio CI scope are pulled by `generate_non_factory_sboms.py`; the script uses an image reference as it's defined in `docker-compose.yml`, so there is no issue with such images regardless of reference type (tag or sha256).

Let me check the case when an image hosted in hub.f.io is referenced by sha in `docker-compose.yml`.

mike-sul: Indeed, the current code allows using the `sha` reference to images built on our CI.
Therefore, if we merge this PR, then our CI will fail in the publish run because the layers metadata gathering logic won't find the `sha` referenced image's layers data.
I see two options how to fix it:

1. Make `generate_non_factory_sboms.py` pull the `sha` referenced images stored in `hub.f.io`.
2. Don't treat the error in the compose-publish (no layers metadata for an image) as a critical and let the run continue its work. This is fine since, since aklite will fallback to the approximate app layer size calculation.

I incline towards #1 because it also makes the SBOMing work correctly in such the case. Currently, it does not do SBOMing for images hosted in hub.f.io  abd referenced by sha in `docker-compose.yml`

mike-sul: This version works fine for all three cases:
1. A factory image is tagged with `:latest` (replace `latest` with the "build" tag)
2. A factory image is not tagged nor pinned (append the "build" tag)
3. A factory image is pinned (nothing to do; image is pulled by the sbom function, hence it's present in a local docker store and referenced by the pinned/sha digest, so the layers metadata gathering function can find it.)

mike-sul: @doanac Did you have a chance to look into this?

doanac (approved):

PR #295, opened 2023-07-24
lmp: Pass lmp version to the customize target func

mike-sul:
Add an LmP version to the Target JSON being built by a given CI build run.
This will allow us to easily determine what version of LmP is installed on a device.
We know which Target is installed on a device, so if the target metadata includes an LmP version, then we know the LmP version installed on the device.


mike-sul:
@vkhoroz FYI, this changes adds another field to `targets.json`, i.e. `custom.lmp-ver`. The goal is to know which LmP version is installed on a device.

```
fioctl targets list --raw | jq . | grep lmp-ver
          "lmp-ver": "91",
          "lmp-ver": "91",

```

ricardosalveti (approved):
LGTM, we just need to remember that while this is a good indication of the version used, we require the user to push the lmp-manifest tags into their own factory when they get updated, otherwise it might show an older rev (part of the update script logic, but some might update by hand).

mike-sul:
> we require the user to push the lmp-manifest tags into their own factory when they get updated, otherwise it might show an older rev

Sure, this is a generic issue that impacts `/etc/os-release` value too, not this change specific.

mike-sul:
> @mike-sul our API allows any _unknown_ field for the target's `custom` attribute. But, if you want this new `lmp-ver` field to be validated (if e.g. it is going to be used somehow downstream), it'd be wise to add the validation for it in our API's `_tuf_validators`.

Well, this is part of our API response, so I suppose we should validate it if customers use the API directly.

mike-sul:
@ricardosalveti Do you think we should add and then print to user just a version (e.g. `91`)? Or maybe we should add the extended version, the one that we add to `os-release`, e.g. `VERSION_ID=4.0.11-2028-91`. The extended version includes a kernel version, a CI build number, and a LmP version.

ricardosalveti:
> The extended version includes a kernel version, a CI build number, and a LmP version.

Minor correction, the first value is the Yocto/OE base version (e.g. dunfell, kirkstone, etc). I believe it is more useful than just the LmP version for sure, so +1.

PR #300, opened 2023-08-08
Revert "lmp/bb-build: since lmp v87 the setscene-only is not need any more"

quaresmajose:
This reverts commit 40a0e96a47306e1f8d71722ccd073d6ab3fa751e.

There is still an issue pending that need the build in two steps. We have seen some failed builds with [1] where the bitbake build everything but the build fails at the end.
More investiagation is needed to root cause and fix the issue on the bitbake code base.

[1] `failed with exit code ‘1' - real task will be run instead`

[FFTK-2690]


ricardosalveti (approved):
LGTM, we can continue investigating the issue but this will at least improve the customer experience a bit by avoiding certain failures.

PR #302, opened 2023-08-09
lmp/bb-build: use status and call bitbake directly

quaresmajose:
The run function have significant overhead because it used the indent which in turn uses sed on every line.
So use the function status to show the build steps and call bitbake directly.


ricardosalveti (approved):
LGTM

PR #303, opened 2023-08-10
Generate and upload delta stats

mike-sul:
- Some refactoring and fixes required by the new functionality.
- Obtain and upload statistics about generated static deltas.
- Update each relevant Target with a reference to the generated and uploaded delta statistic file.

### DIFF
--- a/lmp/generate-static-deltas
+++ b/lmp/generate-static-deltas
@@ -51,8 +51,8 @@ def drain(progress: Progress, response, prog_cb: ProgressCb, dst: str):
 
 
 def _download_extract(progress: Progress, tarurls: List[str], dst: str):
-    status(f"Downloading: {tarurls} -> {dst}ostree_repo")
-
+    status("Downloading the following {} ostree repos into the single one -> {}ostree_repo\n\t{}"
+           .format(len(tarurls), dst, "\n\t".join(tarurls)))
     total_length = 0
     responses = []
     for u in tarurls:
--- a/lmp/generate-static-deltas
+++ b/lmp/generate-static-deltas
@@ -8,7 +8,7 @@ from tempfile import mkstemp
 import requests
 from shutil import rmtree
 from subprocess import Popen, PIPE
-from typing import List, Tuple
+from typing import List
 
 from helpers import (
     Progress,
@@ -19,7 +19,7 @@ from helpers import (
     require_env,
     require_secrets,
     secret,
-    secret_get,
+    http_get,
     status,
 )
 from static_deltas import Delta, generate_deltas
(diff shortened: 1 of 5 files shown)


lmp/generate-static-deltas
```diff
@@ -104,18 +104,18 @@ def main(creds_zip_file: str, deltas: List[Delta], factory: str, tok_secret_name
...
-    pulls = []
+    downloads = set()  # a set of URLs to ostree repo archives to download from Jobserv
+    pulls = set()  # a set of ostree commit hashes to pull from ostreehub (if no URL to archive)
     for d in deltas:
-        downloads.append(d.to[1]) if d.to[1] else pulls.append(d.to[0])
-        work += 1
+        (e, s) = (d.to[1], downloads) if d.to[1] else (d.to[0], pulls)
+        if e not in s:
+            work += 1
+        s.add(e)
```
doanac, line 126: i think you could put `s.add(e)` in the "if" block for a negligible performance improvement.

lmp/generate-static-deltas
```diff
@@ -39,19 +40,19 @@ class ProgressCb:
...
-            prog_cb(len(chunk))
-    p.stdin.close()
-    p.wait()
+    last_pos = 0
+    with io.BufferedReader(r.raw, buffer_size=1024 * 1024) as br:
+        with tarfile.open(fileobj=br, mode="r|bz2") as tar_stream:
+            for member in tar_stream:
+                tar_stream.extract(member, dst)
+                prog_cb(br.tell() - last_pos)
+                last_pos = br.tell()
```
doanac, line 61: i can't remember why i used Popen here. I think i was getting some streaming issues, but maybe that was from the connection stuff you fixed in the other commit.

lmp/static_deltas.py
```diff
@@ -49,6 +52,29 @@ def save_delta_stats(delta_stats: dict, out_dir: str):
...
+                         },
+                         data=s["canonical-json"])
+        if not r.ok:
+            raise requests.exceptions.HTTPError("Failed to upload delta stats to ostreehub; "
+                                                f"{r.status_code}, err: {r.text}")
+        # make sure the `sha` and `size` received from ostreehub matches with the original one
+        if r.json()["sha256"] != s["sha256"]:
+            raise Exception("invalid content hash is received from ostreehub; "
+                            f"expected: {s['sha256']}, received: {r.json()['sha256']}")
+        if int(r.json()["size"]) != len(s["canonical-json"]):
```
doanac, line 74: should we convert size to int? ie if its not an int we should probably fail just do to a bad assertion in our data types.

lmp/static_deltas.py
```diff
@@ -75,6 +76,31 @@ def upload_delta_stats(factory: str, delta_stats: dict, tok_secret_name: str):
...
+    with open(targets_file) as f:
+        targets = json.load(f)
+
+    for target_name, target_value in targets["targets"].items():
+        to_hash = target_value["hashes"]["sha256"]
+        if to_hash in delta_stats:
+            target_value["custom"]["delta-stats"] = {
+                "sha256": delta_stats[to_hash]["sha256"],
+                "size": len(delta_stats[to_hash]["canonical-json"])
+            }
```
doanac, line 93: so we set the sha256 and size and aklite will know how to download this as content addressable and do a limited read of "size" to pull down this delta info?

doanac (commented):
nice cleanup. pretty much ready to go

lmp/generate-static-deltas
mike-sul: I did so in the first place and then thought that you would ask to move it out of `if` since that `if` is not needed for "unification" of the input list, just for the "work" counting. So, if the work counting is moved or removed then algorithm would still work.

lmp/generate-static-deltas
mike-sul: Maybe. I checked the given implementation for the same cases when the current one fails and it works, and works faster.

lmp/static_deltas.py
mike-sul: Won't this `if int(r.json()["size"])` raise an exception if `r.json()["size"]` is not an integer before making comparison.

Actually, I realize now, that I should have made ostreehub return an integer as a value of the "size" field. I don't know why I made it a string in the first place. I will update ostreehub at first, and then this code, so  `r.json()["size"]` becomes an integer.

lmp/static_deltas.py
mike-sul: Exactly, https://github.com/foundriesio/aktualizr-lite/blob/d13145f9b3b65897f3668ca0c2255c042b510904/src/rootfstreemanager.cc#L267

doanac (approved):
LGTM once you get the backend changes inline for this

PR #304, opened 2023-08-17
lmp/jobserv: add stm32mp15-disco-sec target

igoropaniuk:
Add STM32MP15 Discovery Kit (secure boot) target.


ricardosalveti (approved):
LGTM

PR #306, opened 2023-08-31
lmp:bb-config: Deprecate the variable UBOOT_SIGN_ENABLE from factory

angolini:
Setting this variable in local.conf does not take effect for several machines, and the preferable way to set this would be in the meta-subscriber-overrides metadata.


vanmaegima:
@angolini thanks for the investigation and for this change.
From: the preferable way to set this would be in the meta-subscriber-overrides metadata.
This is not ideal, it would be great if we had some kind of overrides for the SoC in meta-lmp.

vanmaegima (approved):

MrCry0 (approved):

PR #307, opened 2023-09-05
lmp: factory-containers: Deprecate ENABLE_PTEST

angolini:
Deprecate the variabel ENABLE_PTEST as ptest is disabled by default since [1]

[1] https://github.com/foundriesio/meta-lmp/commit/2e51562a43522cdce418bba0784cc6dfa91d1d61


ricardosalveti (approved):
LGTM, I don't think we have anyone using it, but it would be good to go over our factories and check.

quaresmajose (approved):

MrCry0 (approved):

PR #308, opened 2023-09-11
lmp/bb-build: add trap to do some pending operations on exit

quaresmajose:
It will be useful to get the buildstats results when OOM kill the bitbake and in any general errors.

### DIFF
--- a/lmp/bb-build.sh
+++ b/lmp/bb-build.sh
@@ -1,5 +1,36 @@
 #!/bin/bash -e
 
+function finish() {
+    # we need to check that because it is not available before kirkstone
+    if command -v bitbake-getvar >/dev/null 2>&1; then
+        # get buildstats path
+        BUILDSTATS_PATH="$(bitbake-getvar --value TMPDIR | tail -n 1)/buildstats"
+    fi
+    # check if the buildstats was enabled
+    if [ -d "$BUILDSTATS_PATH" ]; then
+        # get the most recent folder
+        BUILDSTATS_PATH="$(ls -td -- $BUILDSTATS_PATH/*/ | head -n 1)"
+        # we need to check that because it can't be available in old containers
+        if command -v xvfb-run >/dev/null 2>&1 ; then
+            # producing bootchart.svg
+            run xvfb-run $(realpath $BUILDDIR/../layers/openembedded-core/scripts/pybootchartgui/pybootchartgui.py) \
+                --minutes --format=svg --output=${archive}/bitbake_buildchart $BUILDSTATS_PATH
+        fi
+        # write a summary of the buildstats to the terminal
+        BUILDSTATS_SUMMARY="$(realpath $BUILDDIR/../layers/openembedded-core/scripts/buildstats-summary)"
+        # we need to check that because it is only available in the kirkstone branch
+        if [ -f $BUILDSTATS_SUMMARY ]; then
+            # common arguments with bold disabled
+            BUILDSTATS_SUMMARY="$BUILDSTATS_SUMMARY --sort duration --highlight 0"
+            # log all task
+            $BUILDSTATS_SUMMARY $BUILDSTATS_PATH > ${archive}/bitbake_buildstats.log
+            # for console hide tasks < Seconds
+            BUILDSTATS_SUMMARY="$BUILDSTATS_SUMMARY --shortest 60"
+            run $BUILDSTATS_SUMMARY $BUILDSTATS_PATH
+        fi
+    fi
+}
+
 HERE=$(dirname $(readlink -f $0))
 source $HERE/../helpers.sh
 require_params IMAGE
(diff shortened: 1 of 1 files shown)


lmp/bb-build.sh
```diff
@@ -1,5 +1,36 @@
...
+        BUILDSTATS_PATH="$(bitbake-getvar --value TMPDIR | tail -n 1)/buildstats"
+    fi
+    # check if the buildstats was enabled
+    if [ -d "$BUILDSTATS_PATH" ]; then
+        # get the most recent folder
+        BUILDSTATS_PATH="$(ls -td -- $BUILDSTATS_PATH/*/ | head -n 1)"
+        # we need to check that because it can't be available in old containers
+        if command -v xvfb-run >/dev/null 2>&1 ; then
+            # producing bootchart.svg
+            run xvfb-run ../layers/openembedded-core/scripts/pybootchartgui/pybootchartgui.py \
```
MrCry0, line 16: Wouldn't it be better to use TOPDIR here instead of relative addressing? `$TOPDIR/../layers/...`

lmp/bb-build.sh
```diff
@@ -1,5 +1,36 @@
...
+                --minutes --format=svg --output=${archive}/bitbake_buildchart $BUILDSTATS_PATH
+        fi
+        # write a summary of the buildstats to the terminal
+        BUILDSTATS_SUMMARY="../layers/openembedded-core/scripts/buildstats-summary"
```
MrCry0, line 20: and here

lmp/bb-build.sh
quaresmajose: `TOPDIR` only exist inside the bitbake environment, what we have on bash shell is the `BUILDDIR` that points to the build directory.
Anyway because the `BUILDDIR` was allways created inside the `OEROOT` the layers path is also allways rechable using the `../layers/`
https://github.com/foundriesio/lmp-manifest/blob/427d603fdf5703fee7922509cf947f72f72cda74/setup-environment-internal#L123-L127

MrCry0: I see. I meant `bitbake-getvar --value TOPDIR`, not the env var :)

But if BUILDDIR is available directly, it would be better to use `$BUILDDIR/../layers/...`

The point is to avoid using relative paths.

quaresmajose: Fixed. thanks

MrCry0 (approved):

lmp/bb-build.sh
```diff
@@ -13,11 +13,11 @@ function finish() {
         # we need to check that because it can't be available in old containers
         if command -v xvfb-run >/dev/null 2>&1 ; then
             # producing bootchart.svg
-            run xvfb-run ../layers/openembedded-core/scripts/pybootchartgui/pybootchartgui.py \
+            run xvfb-run $(realpath $BUILDDIR/../layers/openembedded-core/scripts/pybootchartgui/pybootchartgui.py) \
```
angolini, line 16: I don't see a reason for not squashing it to the previous commit.

angolini (approved):

lmp/bb-build.sh
ricardosalveti: @quaresmajose I would probably do the other way around, have this patch first before moving the content to the finish function and then the other patch on top that basically just moves the same content but to a function.

This way it is easy to compare and isolate the changes.

lmp/bb-build.sh
```diff
@@ -21,38 +52,13 @@ bitbake -e ${IMAGE} > ${archive}/bitbake_image_env.txt
 status "Run bitbake (setscene tasks only)"
 bitbake --setscene-only ${IMAGE} || true
 
+# add trap to do some pending operations on exit
+trap finish TERM INT EXIT
```
ricardosalveti, line 56: And at this commit please add a statement saying that this shouldn't imply in any functional changes, it is just moving code around and making it being called on trap.

quaresmajose: we need to return the same exit code as before, also added now inside the trap

lmp/bb-build.sh
quaresmajose: makes sense

ricardosalveti (approved):
LGTM, just needs testing on a factory (testing working and failing scenarios) before we deploy just to make sure.

quaresmajose:
Tested the failing scenario and it works as expected:
```
Summary: 1 task failed:
  /srv/oe/build/conf/../../layers/meta-lmp/meta-lmp-base/recipes-kernel/linux/linux-lmp_6.1.bb:do_compile
Summary: There were 3 ERROR messages, returning a non-zero exit code.
== 2023-09-21 11:40:15 Running: xvfb-run /srv/oe/layers/openembedded-core/scripts/pybootchartgui/pybootchartgui.py --minutes --format=svg --output=/archive/bitbake_buildchart /srv/oe/build/tmp-lmp/buildstats/20230921113638/
|  /srv/oe/layers/openembedded-core/scripts/pybootchartgui/pybootchartgui/draw.py:706: SyntaxWarning: "is not" with a literal. Did you mean "!="?
|    if (OPTIONS.show_pid or OPTIONS.show_all) and ipid is not 0:
|  /srv/oe/layers/openembedded-core/scripts/pybootchartgui/pybootchartgui/draw.py:804: SyntaxWarning: "is not" with a literal. Did you mean "!="?
|    if i is not 0:
|  bootchart written to '/archive/bitbake_buildchart.svg'
|--
Script completed with error(s)
== 2023-09-21 11:40:25.341426: Finding artifacts to upload
```

The `bitbake_buildchart.svg` will be on the root folder of the artifacts and not on the `others/bitbake_buildchart.svg`

ricardosalveti:
> Tested the failing scenario and it works as expected:

And did the CI job failed as expected as well? Thinking on the CI state machine.

ricardosalveti:
I expect so because of 'Script completed with error(s)', but just checking before we merge to be safe :-)

quaresmajose:
Yup, the CI job fails 

image

PR #311, opened 2023-10-24
lmp/bb-config: use variables instead of the content

quaresmajose:
Let's let bitbake take care of the use of the variables:

```
LMP_FACTORY_IMAGE
OSTREE_BRANCHNAME
GARAGE_TARGET_VERSION
GARAGE_TARGET_URL
DISTRO_VERSION
```

All of then have references for other variables:

```
MACHINE
H_BUILD
LMP_VERSION
```

_`BB_ENV_PASSTHROUGH` is not needed in this case because all the varaibles needed is already defined in local.conf.
The main reason for this change is to let bitbake track and show what have changed in the cache signatureres._

_For example:
When the `H_BUILD` change, bitbake will show diferences in all variable that is using the `H_BUILD` somehow and root cause of the cache signature change it's the change of all the afected variables:_

```
H_BUILD
GARAGE_TARGET_VERSION
GARAGE_TARGET_URL
DISTRO_VERSION
```

_With this patch it will show that there are differences in all the variables as in previoes but the root cause of the cache signature change it's just the variable `H_BUILD`._


lmp/bb-config.sh
```diff
@@ -195,14 +195,17 @@ fi
 
 # Add build id H_BUILD to output files names
 if [ "$CONF_VERSION" == "1" ]; then
-	cat << EOFEOF >> conf/local.conf
-DISTRO_VERSION_append = "-\${H_BUILD}-\${LMP_VERSION}"
+	cat << 'EOFEOF' >> conf/local.conf
```
ricardosalveti, line 198: Why changing this line?

lmp/bb-config.sh
```diff
@@ -195,14 +195,17 @@ fi
...
 EOFEOF
 else
-	cat << EOFEOF >> conf/local.conf
-DISTRO_VERSION:append = "-\${H_BUILD}-\${LMP_VERSION}"
+	cat << 'EOFEOF' >> conf/local.conf
+DISTRO_VERSION:append = "${DISTRO_VERSION_EXTENDED}"
 EOFEOF
 fi
+cat << 'EOFEOF' >> conf/local.conf
+DISTRO_VERSION_EXTENDED ?= "-${H_BUILD}-${LMP_VERSION}"
```
ricardosalveti, line 207: Just to facilitate the reading process here, can we first define this variable and later used it on the code blocks changed above?

ricardosalveti:
Don't we need to change bitbake to allow these env variables to be consumed as well?

ricardosalveti:
BB_ENV_PASSTHROUGH or similar.

MrCry0:
just a note: to avoid rewriting a default `BB_ENV_PASSTHROUGH` value, it would be better to use `BB_ENV_PASSTHROUGH_ADDITIONS`

quaresmajose:
BB_ENV_PASSTHROUGH is not needed in this case all the varaibles needed is already defined in local.conf.
The main reason for this change to let bitbake track and show what have changed in the cache signatureres.

For example:
When the `H_BUILD` change, bitbake will show diferences in all variable that is using the `H_BUILD` somehow and root cause of the cache signature change it's the change of all the afected variables:
```
H_BUILD
GARAGE_TARGET_VERSION
GARAGE_TARGET_URL
DISTRO_VERSION
```
With this patch it will show that there are differences in all the variables as in previoes but the root cause of the cache signature change it's just the variable `H_BUILD`.

lmp/bb-config.sh
quaresmajose: I will do

quaresmajose:
The main intention is to facilitate debugind issue in the sstate-cache, there is no functionality changes in this PR and this only affects the bitbake cache signatures.

angolini:
> BB_ENV_PASSTHROUGH is not needed in this case because all the varaibles needed is already defined in local.conf. The main reason for this change is to let bitbake track and show what have changed in the cache signatureres.
> 
> For example: When the `H_BUILD` change, bitbake will show diferences in all variable that is using the `H_BUILD` somehow and root cause of the cache signature change it's the change of all the afected variables:
> 
> ```
> H_BUILD
> GARAGE_TARGET_VERSION
> GARAGE_TARGET_URL
> DISTRO_VERSION
> ```
> 
> With this patch it will show that there are differences in all the variables as in previous but the root cause of the cache signature change it's just the variable `H_BUILD`.

Please add this note to the PR description.

angolini (approved):

quaresmajose:
> > BB_ENV_PASSTHROUGH is not needed in this case because all the varaibles needed is already defined in local.conf. The main reason for this change is to let bitbake track and show what have changed in the cache signatureres.
> > For example: When the `H_BUILD` change, bitbake will show diferences in all variable that is using the `H_BUILD` somehow and root cause of the cache signature change it's the change of all the afected variables:
> > ```
> > H_BUILD
> > GARAGE_TARGET_VERSION
> > GARAGE_TARGET_URL
> > DISTRO_VERSION
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > With this patch it will show that there are differences in all the variables as in previous but the root cause of the cache signature change it's just the variable `H_BUILD`.
> 
> Please add this note to the PR description.

done

ricardosalveti (approved):
LGTM, we can land tomorrow morning to avoid possible issues when we're not around.

quaresmajose:
> LGTM, we can land tomorrow morning to avoid possible issues when we're not around.

Please note I don't have privileges to merge this repo

PR #313, opened 2023-11-03
lmp/build: adapt post-build processing to scarthgap

quaresmajose:
**lmp/build: check availability of the image manifest**
- Skip the loop if the disto image manifest not exist
- Fail if the manifest file pattern not exist
- Consider only the symbolic links of the images manifest
- From the licenses deploy dir of the image:
   - Copy the license.manifest, if not possible fail
   - Copy the image_license.manifest
```

| + for img in ${DEPLOY_DIR_IMAGE}/*${MACHINE}.manifest
| ++ basename '/srv/oe/build/deploy/images/intel-corei7-64/*intel-corei7-64.manifest'
| ++ sed -e s/.manifest//
| + image_name='*intel-corei7-64'
| ++ readlink '/srv/oe/build/deploy/images/intel-corei7-64/*intel-corei7-64.manifest'
| ++ sed -e 's/\..*manifest//'
| + image_name_id=
| Script completed with error(s)
```

-----

**lmp/build: don't fail when the license.manifest not found**
There is a bug in oe-core and sometimes the lic folder is empty.
https://bugzilla.yoctoproject.org/show_bug.cgi?id=15394

The intention of this change is to document and be reverted once
the problem is fixed.

### DIFF
--- a/lmp/build.sh
+++ b/lmp/build.sh
@@ -129,13 +129,12 @@ if [ -d ${DEPLOY_DIR}/sources ]; then
 			grep -q "NAME: ${p}$" ${DEPLOY_DIR_IMAGE}/*.manifest || continue
 
 			# Only archive GPL packages (update *GPL* regex for additional licenses)
-			numfiles=`ls ${DEPLOY_DIR}/licenses/${p}/*GPL* 2> /dev/null | wc -l`
-			if [ ${numfiles} -gt 0 ]; then
-				mkdir -p ${DEPLOY_SOURCES}/${p}/source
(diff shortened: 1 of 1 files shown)


quaresmajose:
Validated just with `lmp-base`

ricardosalveti:
What causes the image manifest to not be generated?

quaresmajose:
Good question, it could be that it is generated with a different name but I don't know.
In the lmp-base distro this happens.

quaresmajose:
instead of having a message we can make this an error and so it becomes clearer what happened.

quaresmajose:
I will debug it more to get the reason.

quaresmajose:
I will do all of this scarthgap adapt in a new PR

PR #319, opened 2024-01-25
apps: Retry if image pull fails

mike-sul:
Set `--retry-times` parameter for the `skopeo` command so it will retry image/layer pulling 3 times if it fails.


doanac (approved):
@Tim-Anderson - this should help with the issue you were seeing

PR #322, opened 2024-02-06
lmp/jobserv: drop xilinx machines from main-next

quaresmajose:
We will disable this temporarily to prepare the next LTS. Currently we need to fork most of the layers in main-next and this will reducing maintenance effort and number of build.


ricardosalveti (approved):
LGTM

PR #323, opened 2024-02-16
helpers: Only send auth header for source.fio/factories

doanac:
These are the only repositories that require it. Furthermore, we've update source.fio/lmp-mirrors to redirect to github/lmp-mirrors. The git cloning fails because libcurl will see we have extra headers for source.fio and prompt for headers to send to github.

### DIFF
--- a/helpers.sh
+++ b/helpers.sh
@@ -71,8 +71,8 @@ function repo_sync {
 
 	if [ -f /secrets/git.http.extraheader ] ; then
 		domain=$(echo $GIT_URL | cut -d/  -f3)
-		status "Adding git config extraheader for $domain"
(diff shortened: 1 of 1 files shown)


doanac:
@ricardosalveti - this is what was breaking things when I made the change to do redirects from source.foundries.io/lmp-mirrors -> github.com/lmp-mirrors. We had auth headers specified for anything to source.foundries.io so it would not do the redirect as a security precaution (it instead prompts for a user/password and CI has no stdin).

ricardosalveti (approved):
Interesting, nice catch, LGTM

MrCry0 (approved):

PR #324, opened 2024-02-26
Fetch apps for offline update

mike-sul:
Add ability to fetch apps for offline update in the publish run if enabled in the factory config. The published TUF targets contain reference to the archive with fetched apps.


apps/publish.sh
```diff
@@ -30,13 +30,20 @@ PUSH_TARGETS=${PUSH_TARGETS-true}
 
 MACHINES=${MACHINES-""}
 PLATFORMS=${MANIFEST_PLATFORMS_DEFAULT-""}
+DONT_LOAD_CERTS=${DONT_LOAD_CERTS-""}
```
doanac, line 33: Seems like we should make `load_extra_certs` work better/different and not introduce a new environment variable.

mike-sul: I'll see what I can do.

mike-sul:
@doanac This https://github.com/foundriesio/ci-scripts/pull/324/commits/930dee4daf5006decd55b0b2d70dbe05a24fefa7 is the commit to make the assemble run re-use the previously fetched apps if possible.

assemble.py
```diff
@@ -265,6 +273,52 @@ def fetch_restorable_apps(target: FactoryClient.Target, dst_dir: str, shortlist:
...
+    fetched_apps = None
+    fetched_apps_uri = None
+    if "fetched-apps" in target.json.get("custom", {}):
+        fetched_apps_str = target.json["custom"]["fetched-apps"].get("shortlist")
+        if fetched_apps_str:
+            fetched_apps = set(
+                x.strip() for x in fetched_apps_str.split(',') if x)
+        else:
+            # if `shortlist` is not defined or empty then all target apps were fetched
+            fetched_apps = set(app_name for app_name, _ in target.apps())
```
doanac, line 286: i'm guessing target.apps() is returning a dict. Could probably do this like:
```
fetched_apps = set(target.apps().keys())
```

assemble.py
```diff
@@ -336,6 +390,15 @@ def get_args():
...
                 logger.info('Fetching Restorable Apps...')
+                previously_fetched_apps_uri, previously_fetched_apps \
+                    = check_and_get_fetched_apps_uri(target, args.app_shortlist)
+                if previously_fetched_apps_uri:
+                    target_apps_dir = os.path.join(apps_root_dir, target.name)
+                    logger.info("Fetching the app archive from the publish run; uri:"
+                                f" {previously_fetched_apps_uri}, apps: {previously_fetched_apps}")
+                    get_and_extract_fetched_apps(previously_fetched_apps_uri, args.token,
+                                                 target_apps_dir)
+                    logger.info(f"The fetched app archive is extracted to {target_apps_dir}")
```
doanac, line 401: double checking that we don't need an `else` fetch_restorable_apps

assemble.py
mike-sul: Yep, looks like a slightly better option.

assemble.py
mike-sul: No `else` is needed. A set of apps fetched for offline update in the publish run can differ from a set of apps for preloading.
So, the current implementation reuses the publish run apps only if it is a sub-set of the apps for preloading. So, we still need to call the fetch to pull the diff. If the sets are the same then we could avoid the fetch here, but I decided it's not worth of additional `if previously_fetched_apps == apps_to_fetch then skip fetching` because `skopeo` is smart enough not to pull blobs if they already present locally.

When/If I move `ci-scripts` from `skopeo` to `composectl` then I'll improve this logic, so the publish run apps are re-used even if its set is not subset of apps for preloading. With `skopeo` it's a bit cumbersome to implement (think about shared blobs folder).

apps/fetch.py
```diff
@@ -4,23 +4,41 @@
...
 
-from factory_client import FactoryClient
 from apps.target_apps_fetcher import SkopeAppFetcher
+from helpers import cmd
+
+
+def fetch_target_apps(targets: dict, apps_shortlist: str, token: str, dst_dir: str):
+    apps_fetcher = SkopeAppFetcher(token, dst_dir)
+    apps_fetcher.fetch_apps(targets, apps_shortlist=apps_shortlist, force=True)
+    apps_fetcher.fetch_apps_images(force=True)
```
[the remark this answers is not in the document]
mike-sul: I fixed it.

apps/publish.sh
mike-sul: I am not sure it is critical.

doanac: Its creating a bit of an "API". What's the problem you see with load_extra_certs?

mike-sul: I used to get error when I was running `test_apps_publishing.sh`.
Somehow, I am getting only warning now, so I am removing this change.

```
== 2024-02-29 15:09:01 Loading extra ca certificates
== 2024-02-29 15:09:01 Loading extra ca certificates
WARNING: ca-certificates.crt does not contain exactly one certificate or CRL: skipping
```

mike-sul: I removed the commit.

PR #329, opened 2024-03-22
apps: Convert shortlist string to list

mike-sul:
- Convert the comma-separated list of apps represented as a string and provided as an input parameter to a list of string.

- Define the app shortlist as a list of strings in each function of the app fetcher.

### DIFF
--- a/apps/fetch.py
+++ b/apps/fetch.py
@@ -15,7 +15,7 @@ from factory_client import FactoryClient
 from helpers import cmd
 
 
-def fetch_target_apps(targets: dict, apps_shortlist: str, token: str, dst_dir: str):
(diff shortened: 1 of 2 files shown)


apps/fetch.py
```diff
@@ -15,7 +15,7 @@
 from helpers import cmd
 
 
-def fetch_target_apps(targets: dict, apps_shortlist: str, token: str, dst_dir: str):
+def fetch_target_apps(targets: dict, apps_shortlist: [str], token: str, dst_dir: str):
```
doanac, line 18: I didn't know that syntax worked `[str]`. I guess this is a newer `List[str]`.

doanac (approved):

PR #334, opened 2024-05-15
apps: Add a flag to disable SBOM generation

doanac:


doanac:
@vanmaegima - this seems to work in my factory. You think this will work for the customer request?

apps/build.sh
```diff
@@ -175,10 +175,14 @@ for x in $IMAGES ; do
 
 	echo "Build step $((completed+2)) of $total is complete"
 
-	status "Doing a syft SBOM scan"
-	sbom_dst=/archive/sboms/${ct_base}/${ARCH}.spdx.json
-	mkdir -p $(dirname $sbom_dst)
-	syft ${ct_base}:$TAG-$ARCH -o spdx-json > $sbom_dst
+	if [ -z "$DISABLE_SBOM" ] ; then
```
mike-scott, line 178: @doanac To implement this a user adds this to the `params` block of `containers`?
```
DISABLE_SBOM: "1"
```

doanac: yep. i'm testing that now

doanac: here it is disabled - https://app.foundries.io/factories/andy-corp/targets/77/ 
here it is enabled(default) - https://app.foundries.io/factories/andy-corp/targets/78/

vanmaegima:
@doanac If that is part of the `containers` block, would that impact platform builds as well? I remember preloaded apps inherit the block from containers, so it'd be good to understand the impact there. I believe the request here is to disable SBOMs for containers only, but I'll double check.

quaresmajose:
> This only disables for containers. SBOM generation for Platform is done via Yocto - I'm not sure how you disable that.

The yocto SPDX can be also be disabled but I don't see much benefit in doing it, in the kirkstone branch it also have negative side effects like reducing the sstate-cache reuse.

vanmaegima (approved):

MrCry0 (approved):

doanac:
rebased on master

PR #337, opened 2024-05-28
Handle offline fetch apps edge cases

mike-sul:
1. Add ability to disable apps update for offline update.
    It adds the `fetched-apps.uri=None` value to each target, so the
    target consumer, fioctl and aklite-offline, knows whether
    to consider apps for the target offline update or not.

2. Don't set `fetched-apps.shortlist` field if shortlist is not specified or empty value.    
    - Use target apps and shortlist intersection as the shortlist of apps to
      fetch. Also, set the `fetched-apps.shortlist` value to this
      intersection.
    
    - Yield an error if no apps found to fetch for the given intersection of
      target apps and the specified shortlist.

### DIFF
--- a/apps/fetch.py
+++ b/apps/fetch.py
@@ -15,12 +15,26 @@ from factory_client import FactoryClient
 from helpers import cmd
 
 
-def fetch_target_apps(targets: dict, apps_shortlist: str, token: str, dst_dir: str):
+def fetch_target_apps(targets: dict, apps_shortlist: set[str], token: str, dst_dir: str) -> dict[str, set[str]]:
     apps_fetcher = SkopeAppFetcher(token, dst_dir)
+    fetched_target_apps: dict[str, set[str]] = {}
     for target_name, target_json in targets.items():
-        apps_fetcher.fetch_target(FactoryClient.Target(target_name, target_json),
-                                  [x.strip() for x in apps_shortlist.split(',') if x]
-                                  if apps_shortlist else None, force=True)
+        target = FactoryClient.Target(target_name, target_json)
+        target_apps = set(app for app, _ in target.apps())
+        target_app_shortlist: set[str] = None
+        if apps_shortlist:
+            target_app_shortlist = target_apps.intersection(apps_shortlist)
+
+        if apps_shortlist and not target_app_shortlist:
+            raise Exception(f'No apps found to fetch; target: {target_name};'
+                            f' target apps: {",".join(target_apps)};'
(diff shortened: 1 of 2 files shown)


mike-sul:
@doanac Since Andre has day offs today and tomorrow, could you please look into it.
I've tested it pretty thoroughly and checked the edge cases mentioned in the commit messages.

apps/fetch.py
```diff
@@ -15,12 +15,26 @@
...
+    fetched_target_apps: dict[str, set[str]] = {}
     for target_name, target_json in targets.items():
-        apps_fetcher.fetch_target(FactoryClient.Target(target_name, target_json),
-                                  [x.strip() for x in apps_shortlist.split(',') if x]
-                                  if apps_shortlist else None, force=True)
+        target = FactoryClient.Target(target_name, target_json)
+        target_apps = set(app for app, _ in target.apps())
+        target_app_shortlist: set[str] = None
+        if apps_shortlist:
+            target_app_shortlist = target_apps.intersection(apps_shortlist)
```
doanac, line 26: should we log what got dropped from this intersection so that its easier to detect if someone made a typo?

apps/fetch.py
```diff
@@ -71,20 +74,27 @@ def main(args: argparse.Namespace):
...
-            }
-            if fetched_target_apps[target] and len(fetched_target_apps[target]) > 0:
-                target_json["custom"]["fetched-apps"]["shortlist"] = \
-                                                            ",".join(fetched_target_apps[target])
+        if args.disable:
+            logging.info("Disabling apps for offline update...")
+            for target, target_json in targets.items():
+                target_json["custom"]["fetched-apps"] = {
+                    "uri": None
+                }
```
doanac, line 82: would it make more sense to do:
` target_json["custom"]["fetched-apps"]  = None`
or
` target_json["custom"]["fetched-apps"]  = {}`

apps/fetch.py
mike-sul: Yep, I think it may be useful. It's not necessary has to be type. Customers may have different set of target apps in different branches. I will add the log.

apps/fetch.py
mike-sul: Maybe, it will require a change in aklite though and I've already update the meta-lmp... This is kind of internal interface between ci-scripts and aklite, not user facing stuff.

apps/fetch.py
mike-sul: Updated.

doanac (approved):

PR #338, opened 2024-06-03
apps: Fetch app images if sbom is turned off

mike-sul:
Pull app images if the sbom generation is turned off.
If the sbom generation is turned ON, it pulls all app images to the local docker store. Then, the following script that gather information about app image layers parses the local docker store, therefore it assumes that all app images are present in the store.
If the sbom generation is OFF, then the 3rd party images are not present in the local store, hence they must be fetched to make the app layer metadata generation script work correctly. This is exactly what this change does.

### DIFF
--- a/apps/fetch_app_images.py
+++ b/apps/fetch_app_images.py
@@ -1,0 +1,41 @@
+import argparse
+import logging
+import os
+import subprocess
+
+from apps.compose_apps import ComposeApps
+from apps.apps_publisher import AppsPublisher
+from helpers import status
+
+
+def main(args):
+    status("Fetching app images to the local docker engine store")
+
+    publisher = AppsPublisher(args.factory, "", "")
+    apps = ComposeApps(args.apps_root, quiet=True)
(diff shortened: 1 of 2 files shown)


apps/fetch_app_images.py
```diff
@@ -0,0 +1,41 @@
...
+from apps.apps_publisher import AppsPublisher
+from helpers import status
+
+
+def main(args):
+    status("Fetching app images to the local docker engine store")
+
+    publisher = AppsPublisher(args.factory, "", "")
+    apps = ComposeApps(args.apps_root, quiet=True)
+    publisher.tag(apps, args.tag)
```
doanac, line 16: so the trick here and why we need all this python vs a "docker compose pull"  is because of the tagging complexities, right?

mike-sul: Yes. In addition to that we need to skip the disabled apps and the "image directories" (directories that doesn't have docker-compose.yml).

doanac (approved):

PR #339, opened 2024-06-06
Fix image layer meta gathering

mike-sul:


apps/build.sh
```diff
@@ -17,7 +17,7 @@ HERE=$(dirname $(readlink -f $0))
 require_params FACTORY
 
 export DOCKER_BUILDKIT=1
-BUILDKIT_VERSION="${BUILDKIT_VERSION-v0.10.3}"
+BUILDKIT_VERSION="${BUILDKIT_VERSION-v0.13.1}"
```
doanac, line 20: are we sure this change is needed. If so - can we update the commit message to explain why?

Also - Can we bump this without bumping the docker-dind-ci container image?

mike-sul: No, I am not sure. Let me double check it.

mike-sul:
```
|  #1 DONE 5.5s
|  ERROR: multiple outputs currently unsupported by the current BuildKit daemon, please upgrade to version v0.13+ or use a single output
```

apps/build.sh
```diff
@@ -127,7 +127,7 @@ for x in $IMAGES ; do
 		docker_cmd="$docker_cmd  --no-cache"
 	fi
 
-	docker_cmd="$docker_cmd --push --cache-to type=registry,ref=${ct_base}:${LATEST}-${ARCH}_cache,mode=max"
+	docker_cmd="$docker_cmd --load --output=type=registry,oci-mediatypes=false --provenance=false --cache-to type=registry,ref=${ct_base}:${LATEST}-${ARCH}_cache,mode=max"
```
doanac, line 130: I think we'll have to have some conditional logic here. I'm not sure we want to move every customer to the new container image. So we'll have to temporarily check for the buildkit version or an environment variable or something that can work in both the current mode and this new mode you are working on.

mike-sul: Do you mean something like this https://github.com/foundriesio/ci-scripts/pull/339/commits/50007efc424b62a9cbc6f1975740834f23e39e4e?

A user can control it through

`containers.ci_image` and `containers.params.BUILDKIT_VERSION` variables in factory config.

doanac: I think that commit will do it. Might make a note in the code that its a "temporary" hack until we move everyone to the new buildkit.

as per enabling the feature. I'd say if `container.ci_image` is defined - we honor it. Otherwise, we start some new allow-list type function to start doing it customer by customer. I'm fine converting all trial/free factories to the new approach.

mike-sul: It looks like a change to the source/repoz, this PR won't require any adaptation to it.

doanac: LGTM - just add a comment before this line so we'll remember that its safe to remove some day in the future

mike-sul: I added a comment. However, I am not sure if it is going to be safe to remove it ever as long as we allow users to specify buildkit/buildx versions through the factory parameters I mentioned above.

doanac: its something i can script and figure out some day.

PR #340, opened 2024-07-17
lmp: Support partner mirrors

doanac:
We are introducing the ability for our partner factories to have sstate mirror objects in GCS. This change detects that based on the "forked-from" attribute we use a factory's factory-config.yml that will get set by our git server as a CI run parameter when triggering a build.


doanac:
@mike-scott - this should get your derivative factory going.

mike-scott (approved):
LGTM.  (Partners should not be including any files in their sstate cache that would need a special token or NDA.  That needs to be managed at the individual Factory level)

MrCry0 (approved):
LGTM

quaresmajose (approved):
LGTM

PR #341, opened 2024-08-08
lmp/build: add support to have a pre created /srv/oe dir

quaresmajose:
With this we can make this dir persistent which helps a lot in debugging with the simulator.

```
  "persistent-volumes": {
    "bitbake": "/var/cache/bitbake",
    "oe": "/srv/oe"
  },
```

To do that we need to inject this line on the `rundef.json` before launching the `jobserv_runner.simulator`:

```
cat <<< $(jq '."persistent-volumes"."oe"="/srv/oe"' rundef.json) > rundef.json
```

### DIFF
--- a/lmp/build.sh
+++ b/lmp/build.sh
@@ -10,6 +10,7 @@ start_ssh_agent
 git_config
 load_extra_certs
 
+mkdir -p /srv/oe
 if [[ $GIT_URL == *"/lmp-manifest.git"* ]]; then
 	status "Build triggered by change to lmp-manifest"
 	manifest="file://$(pwd)/.git -b $GIT_SHA"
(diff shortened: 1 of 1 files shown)


lmp/build.sh
```diff
@@ -36,7 +37,7 @@ cp /root/.netrc /home/builder/.netrc || true
 
 set_base_lmp_version
 
-mkdir build conf
+mkdir -p build conf
```
MrCry0, line 40: -p forces to create mutli-level subdirs. There are 2 separate 1-level subdirs. I doubt we need there `-p`

with this - lgtm

MrCry0 (approved):

lmp/build.sh
doanac: it also won't err out if the directory exists. I think he's doing this instead `[ -d build ] || mkdir build`.

doanac (approved):

lmp/build.sh
quaresmajose: > it also won't err out if the directory exists. I think he's doing this instead [ -d build ] || mkdir build.

yes that was really the reason

MrCry0: > it also won't err out if the directory exists. I think he's doing this instead `[ -d build ] || mkdir build`.

Oh, suck a good trick, didn't aware of this side-effect :)

ricardosalveti (approved):
LGTM

PR #342, opened 2024-08-22
helpers: Support minor releases of LmP in mirror code

doanac:
We are about to release a 94.1 version of the LmP. In the event of a 94.1 tag, we want to truncate the value and still use the "94" cache. This also allows the development of "95" to continue correctly.

The drawback to this approach is that the .1 part of the sstate cache is actually in the 95 bucket. 94.1 is small enough this doesn't matter. If we ever have a minor release that causes sufficient cache invalidation, then we'll need to re-think how to produce the cache better.


quaresmajose (approved):
LGTM

PR #346, opened 2024-09-19
apps/build.sh: Make dockerd listen only on unix socket

mike-sul:
Makes the `dockerd-entrypoint.sh` script to run dockerd in the way it listens only the usual unix socket and avoiding listening on network ports `2375` or `2376`.
Also, it seems important to run the docked through `/usr/local/bin/dind`.


mike-sul:
This change is agnostic to the dind version change.

doanac (approved):
LGTM

PR #347, opened 2024-09-19
Enable oci image format

mike-sul:


doanac (commented):
LGTM - I do think this is something we'll want to somehow document so that customer support can explain what you are about to do.

mike-sul:
> LGTM - I do think this is something we'll want to somehow document so that customer support can explain what you are about to do.

Sure. I don't think we should enable OCI image format for all factories by default at once. At first, let's update dind-ci image gradually (dev, some factories, all factories). 
In addition to that we should make OCI image enabling configurable, maybe just another factory/build parameter.

mike-sul:
@doanac It can be as simple as https://github.com/foundriesio/ci-scripts/pull/347/commits/1b4884a8938c2f2ce458abfa652f29f6eb430bed.

doanac (approved):
seems clean enough to me

mike-scott:
We're missing commit descriptions which explain why we're making these changes.

mike-sul:
> We're missing commit descriptions which explain why we're making these changes.

Updated the commit messages.

mike-scott:
> Updated the commit messages.

Thanks @mike-sul 
LGTM from a commit description PoV

PR #348, opened 2024-09-20
apps/build.sh: Print detected arch in runner container

mike-sul:
It can be useful to see what machine architecture the `arch` detects inside of the CI container.


doanac (approved):
I was today years old when I learned about this command and that it's a wrapper for `uname -m`.

PR #350, opened 2024-09-24
lmp/jobserv: remove the TI machines on the scarthgap-next

quaresmajose:
For the v95 scarthgap based the meta-ti layer are not compatible with the scarthgap-next oe-core fork we currently have but it builds without it well. So I will remove the TI layer and machines from CI job because it brings no value to continue to exercise this than always fails.


angolini (approved):

quaresmajose:
The oe-core fork should be dropped soon

PR #359, opened 2024-10-25
lmp/bb-build: only use the sstate-cache mirror on the first step

quaresmajose:
On the fisrt build step we run bitbake for setscene only tasks and obvious this one is the only one that use the reads and use the sstate-cache.
The second step  will be a 0% hit of available because everything has already been accelerated in the first step. In this second step everything will be build from scratch and the local sstate-cache populadted with the artifacts.

This will improve and significantly reduce the load generated on the sstate-mirror http server by our CI builds.


angolini (commented):
there are 2 typos on the commit log
fisrt
populadted

The description is confusing to me, maybe, instead of saying "first step" and "second step", say the command? Because there are other bitbake steps, so which you mean by first?

quaresmajose:
I have updated the commit, hope it is more clear now
```
    lmp/bb-build: only use the sstate-cache mirror when it is necessary
    
    Our build can be divided into two steps:
    1 - build only using the sstate-cache
    2 - build everything else missing
    
    On the step [1] we run bitbake for setscene only tasks
    and obvious this one is the only one that uses and reads the
    sstate-cache.
    
    The step [2] will have 0% hit of cache available because everything
    has already been accelerated in the step [1]. In the step [2]
    everything will be build from scratch and the local sstate-cache
    will be populated with the new artifacts.
    
    This change will improve and significantly reduce the load generated on
    the sstate-mirror http server by our CI builds.

```

ricardosalveti:
@quaresmajose I understand the reduced load on the server side, but did you measure if this also have any impact on the build time during the CI run?

quaresmajose:
I am testing this on the scarthgap factory but I didn't measure the time.
The time should be very similar because all the bitbake steps are the same except for the initial step where we ask the http server if it has the artifact xyz. When we are compiling the code this ask to http server, at the beginning of the build, is not needed and the answer given by the server is already known and is no.

quaresmajose:
In theory the build time should be about 2 minutes less, which is the time we waste asking the http server.

ricardosalveti (approved):
LGTM

ndechesne:
btw, in our CI, don't we have access to the sstate mirror through NFS by any chance?

quaresmajose:
> btw, in our CI, don't we have access to the sstate mirror through NFS by any chance?

The internal sstate mirror of the factory (every factory has its own) works through NFS but the public only in HTTPS

igoropaniuk (approved):
LGTM

MrCry0 (approved):

PR #363, opened 2024-11-15
apps: Remove ref to parent target apps if app target

mike-sul:
If a target is produced by a container CI job then there is no need in having `target.custom.origUriApps` value in the target since `target.custom.uri` refers to a CI job that produced it. The `origUriApps` should be set only by the LmP target producer to refer to the most recent container CI job that produced apps inherited by a given LmP target.


MrCry0 (approved):

detsch (approved):
LGTM

PR #364, opened 2024-11-15
helpers: repo_sync: don't use the operator &&

quaresmajose:
Because we run the script with set -e
Any return value other than 0 causes the process to abort.

Using the operator && causes the script to abort because when the command fails there is no path to follow.
If we use the if clause operator we solve the problem.


helpers.sh
```diff
@@ -76,16 +76,22 @@ function repo_sync {
 	fi
 	_repo_extra_args=""
 	for i in $(seq 4); do
-		run repo init $_repo_extra_args --repo-rev=v2.35 --no-clone-bundle -u $* ${REPO_INIT_OVERRIDES} && break
+		if run repo init $_repo_extra_args --repo-rev=v2.35 --no-clone-bundle -u $* ${REPO_INIT_OVERRIDES}; then
```
MrCry0, line 79: There is another way to always return true in such commands:
`		run repo init $_repo_extra_args --repo-rev=v2.35 --no-clone-bundle -u $* ${REPO_INIT_OVERRIDES} && break || true`

What do you think about this way?

quaresmajose: using `if` for this looks better to me and is more clean

doanac (approved):

igoropaniuk (approved):

angolini (approved):

PR #367, opened 2024-12-10
helpers: Use shallow clones for android repo

doanac:
We don't need full repo histories when syncing. This will make CI a little faster and less error prone.


Tim-Anderson (approved):
LGTM

PR #368, opened 2024-12-10
helpers: Use shallow clones for android repo

doanac:
We don't need full repo histories when syncing. This will make CI a little faster and less error prone.


quaresmajose:
Another option could be to save to the persistent space of factory where we also have the download and sstate cache

MrCry0 (approved):
LGTM

doanac:
> Another option could be to save to the persistent space of factory where we also have the download and sstate cache

True. But the less we require of persistent storage - the better.

angolini:
What is the status of this PR?

doanac:
I think we should just merge. I think @caiotpereira - was wanting to somehow migrate customers to it.

PR #370, opened 2025-01-13
lmp/build: Improve license and source-code handling

quaresmajose:
**- License handling**

Instead of blindly failing, this will improve the situation and will show up what is the missing manifests.

- Skip lmp-mfgtool disto variant ()
- Fail if the image manifest not exist
- Consider only the symbolic links of the images manifest
- From the licenses deploy dir of the image:
 - Copy the license.manifest, if not possible fail
 - Copy the image_license.manifest

_For the lmp-mfgtool we don't produce an image and don't make sense to processes the image manifest._

The following is an example of such blindly CI build fail:
```
| + for img in ${DEPLOY_DIR_IMAGE}/*${MACHINE}.manifest
| ++ basename '/srv/oe/build/deploy/images/intel-corei7-64/*intel-corei7-64.manifest'
| ++ sed -e s/.manifest//
| + image_name='*intel-corei7-64'
| ++ readlink '/srv/oe/build/deploy/images/intel-corei7-64/*intel-corei7-64.manifest'
| ++ sed -e 's/\..*manifest//'
| + image_name_id=
| Script completed with error(s)
```

**- Source-code handling**

Adding support for collecting duplicated sources

On multiconfig machines the sources and the license can be duplicated.
We have two places with the source-code and licenses on the ti bsp [1].
If it was the case we will deploy only one package, we will search for the first package license path and use that.

We can also escape and jump this source-code collecting step when we not have any IMAGE_MANIFESTS (i.e no image produced).

[1] https://github.com/foundriesio/meta-lmp/blob/7657b41046ff745ce5fe61a64f615623ebb098e0/meta-lmp-bsp/conf/machine/include/lmp-machine-custom.inc#L457-L459

### DIFF
--- a/lmp/build.sh
+++ b/lmp/build.sh
@@ -101,7 +101,9 @@ for img in ${IMAGE_MANIFESTS}; do
 		ln -sf ${image_name_id}.license.manifest ${DEPLOY_DIR_IMAGE}/${image_name}.license.manifest
 	else
 		status "Image ${image_name} license manifest not found on ${DEPLOY_DIR}/licenses, license manifest can't be collected"
-		exit 1
+		# FIXME: there is a bug in oe-core and sometimes the lic folder is empty
+		# https://bugzilla.yoctoproject.org/show_bug.cgi?id=15394
+		#exit 1
 	fi
 	# Also take care of the image_license, which contains the binaries used by wic outside the rootfs
 	if [ -f ${image_path}/image_license.manifest ]; then
--- a/lmp/build.sh
+++ b/lmp/build.sh
@@ -81,18 +81,31 @@ rm -f ${DEPLOY_DIR_IMAGE}/*.txt
 ## Only publish wic.gz
 rm -f ${DEPLOY_DIR_IMAGE}/*.wic
 
-status "Post-build processing (license)"
+IMAGE_MANIFESTS=""
+if [ "${DISTRO}" != "lmp-mfgtool" ]; then
+	status "Post-build processing (license)"
+	# find image manifests and only consider symbolic links
+	IMAGE_MANIFESTS=$(find ${DEPLOY_DIR_IMAGE} -type l -name "*${MACHINE}*.manifest")
+	if [ "${IMAGE_MANIFESTS}" = "" ]; then
+		status "Image manifest not found on ${DEPLOY_DIR_IMAGE}, license manifest can't be collected"
+		exit 1
+	fi
+fi
 # Link the license manifest for all the images produced by the build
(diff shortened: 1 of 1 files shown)


quaresmajose:
This is not required for the scarthgap version but I discovered this when I was working on that-

lmp/build.sh
```diff
@@ -82,16 +82,32 @@ rm -f ${DEPLOY_DIR_IMAGE}/*.txt
 rm -f ${DEPLOY_DIR_IMAGE}/*.wic
 
 # Link the license manifest for all the images produced by the build
-for img in ${DEPLOY_DIR_IMAGE}/*${MACHINE}.manifest; do
+for img in ${DEPLOY_DIR_IMAGE}/*${MACHINE}*.manifest; do
```
ricardosalveti, line 85: Probably better to change to something like 'for img in `find ${DEPLOY_DIR_IMAGE} -name....`' instead, and handle the use case when manifest is not available outside this for.

lmp/build.sh
```diff
@@ -82,16 +82,32 @@ rm -f ${DEPLOY_DIR_IMAGE}/*.txt
...
+	if [ "${DISTRO}" = "lmp-mfgtool" ]; then
+		status "Image manifest not exist in this distro, skipping"
+		break
+	fi
+	if ! [ -e "${img}" ]; then
+		status "Image manifest not found on ${DEPLOY_DIR_IMAGE}, license manifest can't be collected"
+		eixt 1
```
ricardosalveti, line 92: typo, eixt

lmp/build.sh
quaresmajose: With find it's cleaner and the handle when manifest is not available can right after the finding. I will retest this in my factory

quaresmajose: done

ricardosalveti (approved):
LGTM, we just won't have a way to notify our customer of such missing file at this stage, we need to work on getting the possible fix backported to scarthgap.

quaresmajose:
> LGTM, we just won't have a way to notify our customer of such missing file at this stage, we need to work on getting the possible fix backported to scarthgap.

I will try to see how to do this.

quaresmajose:
The scarthgap is broken with this and I need validate this again on both scarthgap + kirkstone

quaresmajose:
> The scarthgap is broken with this and I need validate this again on both scarthgap + kirkstone

Validated on both scarthgap + kirkstone ci based builds.

ricardosalveti (approved):
LGTM, thanks

PR #371, opened 2025-01-15
lmp/jobserv: prepare for kirkstone/scarthgap/main builds

quaresmajose:
Our main release CI job called `build-release` is tracking the `lmp-manifest/main` branch which is already based on `kirksotne` branches for all the layers. So it is time to rename that job to `build-kirkstone` and also start tracking the `lmp-manifest/kirksotne` branch that we already have.

This will also remove the old `build-release-stable` job which has not been used for some time now and starts to be problematic because it tracks the `kirkstone` branch that we didn't have but we recently created.

After this PR we will have one job used for develpment:

- build-main

And more two for the corresponding LTS branches:

- build-kirkstone
- build-scarthgap

_After the final revision it was agreed to renamed jobs to:_

**build-eol** _kirkstone branch_
**build-lts** _scarthgap branch_
**build-main** _main branch_


angolini:
Can I understand we are going to perform 3 different kind of builds? main, kirkstone and scarthgap?

Also, can you, please, describe how the test builds is implemented in a section (or child page) in the confluence page for Release Process?

quaresmajose:
> Can I understand we are going to perform 3 different kind of builds? main, kirkstone and scarthgap?

right, we will have 3 jobs:

- build-kirkstone
- build-scarthgap
- build-main

I think this naming is more appropriate as it makes clear what branch form the Yocto LTS we are using.

> 
> Also, can you, please, describe how the test builds is implemented in a section (or child page) in the confluence page for Release Process?

I can add something but I don't fully understand the tests.

quaresmajose:
We have already the kirkstone branch created on lmp-manifest and meta-lmp but the scarthgap need to be created on the two repos as well.

lmp/jobserv.yml
```diff
@@ -299,129 +299,6 @@ triggers:
         persistent-volumes:
           bitbake: /var/cache/bitbake
 
-  - name: build-release-stable
```
ricardosalveti, line 302: This was actually used even as part of our previous kirkstone merge (https://ci.foundries.io/projects/lmp/builds/2753/).

lmp/jobserv.yml
```diff
@@ -1,6 +1,6 @@
 timeout: 540  # a build with no cache is quite slow
 triggers:
-  - name: build-release
+  - name: build-kirkstone
```
ricardosalveti, line 3: The machines for kirkstone were actually aligned with build-release-stable, build-release is actually aligned with scarthgap.

lmp/jobserv.yml
```diff
@@ -122,9 +122,9 @@ triggers:
       GIT_URL: |
         https://github.com/foundriesio/lmp-manifest.git
       GIT_POLL_REFS: |
-        refs/heads/main-next
-      OTA_LITE_TAG: 'main-next:main'
-      AKLITE_TAG: promoted-next
+        refs/heads/main
```
ricardosalveti: At the moment we can only do releases based on main, I believe we decided to first release v95 out of main before switching main to the actual oe-core main, didn't we? @angolini ^

lmp/jobserv.yml
```diff
@@ -126,6 +126,13 @@ triggers:
       OTA_LITE_TAG: 'scarthgap'
       AKLITE_TAG: promoted-scarthgap
     runs:
+      - name: lmp-sdk
+        host-tag: amd64
+        container: foundries/dind-ci:19.03.9_b38f166
```
ricardosalveti, line 131: Why are we using this old tag for lmp-sdk?

lmp/jobserv.yml
quaresmajose: > build-release-stable is currently basically a post merge run for kirkstone.

this build-release-stable is already there before we create the kirkstone branch and the last change of the job is from 3 year ago. before it is used for honister.

> This was actually used even as part of our previous kirkstone merge (https://ci.foundries.io/projects/lmp/builds/2753/).

Right but please note this is a side effect of the kirkstone branch creation. When we created the kirkstone branch we should have changed the configuration of the build-release job that follows refs/heads/main to start following the refs/heads/kirkstone.

lmp/jobserv.yml
quaresmajose: This is what we should have done at the time of creating the branch as I mentioned in the previous comment.
_When we created the kirkstone branch we should have changed the configuration of the build-release job that follows refs/heads/main to start following the refs/heads/kirkstone._

lmp/jobserv.yml
quaresmajose: we only run this lmp-sdk in the build-kirkstone and I copy it also to build-scarthgap and build-main. the tag is the same

lmp/jobserv.yml
quaresmajose: Ok given that we can only do releases based on main , maybe is better to define two following path/branches for before and after v95:

**Before v95:**
build-scarthgap
`refs/heads/main`
build-main
`refs/heads/main-next`

I need to change the current PR to follow this.

**After v95:**
build-scarthgap
`refs/heads/scarthgap`
build-main
`refs/heads/main`

I create a new PR to be merged after v95 is released.

quaresmajose: I mean adding this on top of this PR and reverting it after v95 is released.
```diff
diff --git a/lmp/jobserv.yml b/lmp/jobserv.yml
index 31d0cc0..3164949 100644
--- a/lmp/jobserv.yml
+++ b/lmp/jobserv.yml
@@ -122,7 +122,7 @@ triggers:
       GIT_URL: |
         https://github.com/foundriesio/lmp-manifest.git
       GIT_POLL_REFS: |
-        refs/heads/scarthgap
+        refs/heads/main
       OTA_LITE_TAG: 'scarthgap'
       AKLITE_TAG: promoted-scarthgap
     runs:
@@ -227,7 +227,7 @@ triggers:
       GIT_URL: |
         https://github.com/foundriesio/lmp-manifest.git
       GIT_POLL_REFS: |
-        refs/heads/main
+        refs/heads/main-next
       OTA_LITE_TAG: 'main'
       AKLITE_TAG: promoted-main
     runs:

```

quaresmajose: updated in https://github.com/foundriesio/ci-scripts/pull/371/commits/f637aba38dd650c48e172dded1cec2cd4600064a

mwasilew:
This change should be aligned with the testing setup. At the moment the test scheduler is relying on the build names. So changing `build-release` to `build-main` will break automated merge/build/test setup: https://github.com/foundriesio/conductor/blob/master/conductor/api/views.py#L261

LTS testing is currently tracking `kirkstone` branch in lmp-manifest, but again the name of the build matters.

Alternatively there should be a patch to jobserve to allow for some other field to be added so it can be tracked easily not looking at the build name.

lmp/jobserv.yml
quaresmajose: the reason of this remaining is because we will have more than one type of build-release, for example v94.x kirkstone based and v95+ scarthgap based.

quaresmajose:
> This change should be aligned with the testing setup. At the moment the test scheduler is relying on the build names. So changing `build-release` to `build-main` will break automated merge/build/test setup: https://github.com/foundriesio/conductor/blob/master/conductor/api/views.py#L261

We can keep the names that are already in use but I think we need 3 different types of builds so we need to add one more at least.

> LTS testing is currently tracking kirkstone branch in lmp-manifest, but again the name of the build matters.

I think it would be clearer if the names of the build were the same as the branches used and that was the main reason for naming them:

build-main: follow master branch 
build-kirkstone: follow kirkstone branch
build-scarthgap: follow scarthgap branch

if it is possible to have a name `build-release` following two distinct branches `kirkstone` and `scarthgap` perhaps the process can be simplified.

angolini:
I think the amount of builds we have should be related with the release support strategy we want. I'm not sure having 3 is what we want.
But we can decide it is. So let's think about the naming.
If naming affects the testing mechanism, I don't see why we are changing the name of each build every time a new LTS is out. Anyway, a new LTS is out by 2 years, so it's not that much of effort.

but I would keep the build name different from the yocto LTS version:
build-dev -> main
build-release -> scarthgap (when the time comes)
build-stable? build-stale? -> kirkstone (when the time comes)

quaresmajose:
> but I would keep the build name different from the yocto LTS version:
> build-dev -> main
> build-release -> scarthgap (when the time comes)
> build-stable? build-stale? -> kirkstone (when the time comes)

We can have better names, another suggestion:

-  build-dev -> (main)
-  build-lts -> current LTS (scarthgap)
-  build-eol -> last LTS (kirkstone)

Although we don't have to change the names, we will always have to change the branch used, which changed with each new LTS.

mwasilew:
I like what @angolini proposed the most. From the techincal pov it doesn't matter that much. Once we agree on the names I'll push proper changes to conductor.

lmp/jobserv.yml
quaresmajose: With the launch of v95 I revert the https://github.com/foundriesio/ci-scripts/commit/f637aba38dd650c48e172dded1cec2cd4600064a

quaresmajose:
> Please just drop the patch instead of adding and reverting it as part of the same pr.

done

lmp/jobserv.yml
quaresmajose: I have dropped the job to build the container from the build-kirkstone and build-main and keep it only on the build-scarthgap.

quaresmajose: I will add in another [PR381](https://github.com/foundriesio/ci-scripts/pull/381) the change to build it on build-main but for this I need @doanac help because we need to tag the container with a different name like _next_ or something to only be used on the main branch.

quaresmajose:
> > but I would keep the build name different from the yocto LTS version:
> > build-dev -> main
> > build-release -> scarthgap (when the time comes)
> > build-stable? build-stale? -> kirkstone (when the time comes)
> 
> We can have better names, another suggestion:
> 
> * build-dev -> (main)
> * build-lts -> current LTS (scarthgap)
> * build-eol -> last LTS (kirkstone)
> 
> Although we don't have to change the names, we will always have to change the branch used, which changed with each new LTS.

We still need to define and agree in choosing the CI build names?

lmp/jobserv.yml
quaresmajose: I have droped the two coomits above https://github.com/foundriesio/ci-scripts/commit/f637aba38dd650c48e172dded1cec2cd4600064a and https://github.com/foundriesio/ci-scripts/commit/f637aba38dd650c48e172dded1cec2cd4600064a. We can forget all this path/branches for before and after v95.

mwasilew:
I'd like to avoid using branch names in the build names. Something like "main", "lts", "eol" would be my preference.

ricardosalveti:
> I'd like to avoid using branch names in the build names. Something like "main", "lts", "eol" would be my preference.

Yeah, let's do main, lts and eol.

quaresmajose:
Jobs renamed to:

- build-eol
- build-lts
- build-main

ricardosalveti (approved):

PR #372, opened 2025-01-20
Scarthgap: add temporarily workaround to the github pr job

quaresmajose:
We need this temporarily to get a successful github pr of the scarthgap build

PR #373, opened 2025-01-21
lmp/jobserv: temporarily drop from github_pr the deprecated machines

ricardosalveti:
The deprecated machines were removed from the scarthgap and we will fix it in a followup.


ricardosalveti:
Force merging based on the reviews from https://github.com/foundriesio/ci-scripts/pull/372 as jose is off today.

PR #379, opened 2025-02-12
lmp/build: print a status in each step of the post build processing

quaresmajose:
It will be usefull for debug broken builds


ricardosalveti (approved):
LGTM

PR #382, opened 2025-05-08
lmp/build-sdk-container: override the latest tag via an environment variable

quaresmajose:
If the environment variable is present, use it as a tag when creating the container.
The default is build with the tag latest and also with the corresponding version tag.

### DIFF
--- a/lmp/build-sdk-container.sh
+++ b/lmp/build-sdk-container.sh
@@ -3,7 +3,7 @@
 HERE=$(dirname $(readlink -f $0))
 source $HERE/../helpers.sh
 
-TAG=$(git log -1 --format=%h)
+LATEST=${LATEST:-latest}
 
 status Launching dockerd
 unset DOCKER_HOST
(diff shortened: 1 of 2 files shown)


quaresmajose:
The main intention of this PR is to add the possibility to build `hub.foundries.io/lmp-sdk:next` on the lmp-manifest/main branch.

lmp/build-sdk-container.sh
```diff
@@ -18,11 +16,15 @@ for i in `seq 12` ; do
 done
 
 container="hub.foundries.io/lmp-sdk"
-tagged="${container}:${TAG}"
-latest="${container}:latest"
```
doanac, line 22: I think you'll wind up needing a block like:
 https://github.com/foundriesio/ci-scripts/blob/master/lmp/jobserv.yml#L18-L24

for changes to the "next" branch. Seems like the easiest way to control that will be via an environment variable. e.g.
```
- name: lmp-sdk-next
        host-tag: amd64
        container: foundries/dind-ci:19.03.9_b38f166
        privileged: true
        params:
           LATEST=next
        script-repo:
          name: fio
          path: lmp/build-sdk-container.sh
 ```

And then you could have this script take a variable for "latest". e.g:
```
LATEST=${LATEST:latest}
``

quaresmajose: Well! With the argument it wouldn't work. thanks for the tip.
Also added the block on the main build job:
```
- name: lmp-sdk-next
        host-tag: amd64
        container: foundries/dind-ci:19.03.9_b38f166
        privileged: true
        params:
           LATEST=next
        script-repo:
          name: fio
          path: lmp/build-sdk-container.sh
```

ricardosalveti (approved):

PR #394, opened 2025-06-26
lmp/jobserv: main-next: drop qemuarm64-secureboot

quaresmajose:
The qemuarm64-secureboot is broken because of the fitimage so disable it until we get it working again.


angolini (approved):
ok

PR #406, opened 2026-03-04
apps: Add logic to optionally build things in subdirectories

doanac:
When users want to use something like submodules, they'll wind up with repository structure looking like:

 repo:
    submodule-foo/
                  ui/Dockerfile
		  docker-compose.yml

This change will allow us to build a container named "submodule-foo-ui"


mike-sul (approved):
LGTM

detsch (approved):

PR #408, opened 2026-03-27
helpers: Decide LMP_VERSION by branch instead of HEAD

angolini:
Instead of always searching for the latest tag from main, first detect which branch was initialized by repo and then fetch the latest tag.

# Tested by

```
repo init --no-clone-bundle -u https://github.com/foundriesio/lmp-manifest -b main
export H_PROJECT=lmp
source /home/daiane/foundries/ci-scripts/helpers.sh 
set_base_lmp_version


repo init --no-clone-bundle -u https://github.com/foundriesio/lmp-manifest -b scarthgap
export H_PROJECT=lmp
source /home/daiane/foundries/ci-scripts/helpers.sh 
set_base_lmp_version
```

I'm not exactly sure how to get the first `repo init` command line, so I guessed. If there is a better way to test it, please advice.

It returns 96 for main and 97 for scarthgap, as the main logic of using the *next* version is maintained

### DIFF
--- a/helpers.sh
+++ b/helpers.sh
@@ -115,9 +115,22 @@ function set_base_lmp_version {
 	git remote set-url origin https://github.com/foundriesio/lmp-manifest
 	git fetch origin --tags --quiet
 
-	# 3: Find our base LMP version based on the HEAD
-	export LMP_VERSION=$(git describe --tags --abbrev=0 HEAD)
-	export LMP_VERSION_MINOR="$(git rev-list ${LMP_VERSION}..HEAD --count)"
+	# 3: Find our base LMP version based on the branch
+	# repo always checks out manifests.git on a local branch named 'default'
+	# regardless of the actual remote branch; resolve the real branch via tracking config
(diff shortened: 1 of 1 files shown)


ricardosalveti:
But then main will use the v96 + 1 bucket, is this what we want?

ricardosalveti:
If we don't expect releasing main anymore I would just move main to use a different sstate cache bucket and drop most of this logic.

angolini:
@ricardosalveti I'm still unsure how to proceed, in an attempt of better understanding this I made this silly change.

Not exactly thinking about tags on main, but tags on each LTS branch, how to keep it up with CVE stable updates being released in the future.

I agree that the sstate_cache remote directories management has to align with this logic. But today this management is unclear to me, so I hear any feedback.

angolini:
We have an internal discussion to better align the logic with our strategy, so I'm closing this PR now, and if needed we can take the code from here, or reopen it - as needed

PR #412, opened 2026-07-10
fix(apps): select layer digest for the correct source repository

mike-sul:
The same diff ID may map to different layer digests in different source repositories. Match the source repository when looking up a layer digest by its diff ID.

Add basic unit tests for source repository parsing and layer digest selection.


doanac (approved):
thanks

